PR #23, opened 2022-07-31
Map INT32 Kafka connect to INT256 CH type

subkanthi:
Map INT32 Kafka connect to INT256 CH type


UnamedRus:
Curious, why map Int32 to Int256?

PR #80, opened 2022-09-12
python scripts to run checksums

aadant:


python/db_compare/clickhouse_table_checksum.py
```diff
@@ -0,0 +1,411 @@
...
+
+    external_column_types = ""
+    for column in primary_key_columns:
+        external_column_types += ","+column+" "+data_types[column]
+
+    logging.debug("order by columns "+order_by_columns)
+    return (query, select, order_by_columns, external_column_types)
+
+
+def selectTableStatements(table, query, select_query, order_by, external_column_types):
```
subkanthi, line 233: can we please change this to snake case, so its consistent.

python/db_compare/clickhouse_table_checksum.py
```diff
@@ -0,0 +1,411 @@
...
+""".format(table_schema=table_schema, table_name=table_name)
+    (rowset, count) = executeStatement(sql)
+    res = []
+    for row in rowset:
+        if row[0] is not None:
+            res.append(row[0])
+    return res
+
+
+def getTableChecksumQuery(table):
```
subkanthi, line 144: can we please change this to snake case, so its consistent.

python/db_compare/clickhouse_table_checksum.py
```diff
@@ -0,0 +1,411 @@
...
+        rowset = clickhouse_execute_conn(conn, strSql)
+        rowcount = len(rowset)
+    if len(w) > 0:
+        logging.warning("SQL warnings : "+str(len(w)))
+        logging.warning("first warning : "+str(w[0].message))
+
+    return (rowset, rowcount)
+
+
+def executeStatement(strSql):
```
subkanthi, line 71: can we please change this to snake case, so its consistent.

subkanthi (changes requested):

python/db_compare/clickhouse_table_checksum.py
subkanthi: fill fix this in a separate PR.

PR #168, opened 2022-12-07
Added option in clickhouse checksum utility to connect to secure server

subkanthi:
Added option in clickhouse checksum utility to connect to secure server
Add option to override sign column.


aadant (commented):
LGTM

PR #185, opened 2023-01-17
Use antlr4 mysql workbench grammar to generate the MySQL parser

aadant:
Using  https://github.com/mysql/mysql-workbench.git and antlr-4.11.1-complete.jar to generate the MySQL parser.

https://github.com/mysql/mysql-workbench/tree/8.0/library/parsers/grammars

The parser is used with a tree walker and a listener to translate MySQL create tables to CH SQL.

Please note that the generated python was amended because the grammar is designed for C++, not python.

To test the parser separately you can run this command 

```
python db_load/mysql_parser/mysql_parser.py test1.sql

echo "alter table t add c1 int after c0, add column c2 int not null, add c10 Date null, add c11 int, drop column c3, modify column c4 DateTime(6) not null default '' first,  add key k(c2), rename to t1, rename column c3 to c4, change column c1 c5 int not null, change c5 c6 bigint not null;" >test9.sql

python db_load/mysql_parser/mysql_parser.py test9.sql
```


aadant:
The support for alter needs more work :

- the field definitions that are supported in create table in CH are not supported in the alter.
- MySQL semantics of multiple alters is different from CH and needs special attention

MySQL can support this : adding a column that already exists if it is renamed, CH does not.
```
drop table if exists t;
drop table if exists t1;
create table t(c0 int, c4 DateTime, c5 int, c3 int, c2 int, c1 int) engine=InnoDB;
alter table t add c1 int after c0, add column c2 int not null, add c10 Date null, add c11 int, drop column c3, modify column c4 DateTime(6) not null default '2022-01-01' first, 
 add key k(c2), rename to t1, rename column c2 to c7, change column c1 c5 int not null, change c5 c6 bigint not null;
```

PR #308, opened 2023-10-10
Clickhouse loader fixes

AdamKatzDev:
- strip COMMENT from DDL (Doctrine ORM generates columns like `created_at datetime NOT NULL COMMENT '(DC2Type:datetime_immutable)'`)
- fixed crash in `clickhouse-client` call related to missing password variable
- updated to build schema according to the new replacing merge tree engine

### DIFF
--- a/sink-connector/python/db_load/clickhouse_loader.py
+++ b/sink-connector/python/db_load/clickhouse_loader.py
@@ -410,7 +410,7 @@ def load_data(args, timezone, schema_map, dry_run = False):
         for data_file in data_files:
             # double quote escape logic https://github.com/ClickHouse/ClickHouse/issues/10624
             structure = columns.replace(","," Nullable(String),")+" Nullable(String)"
-            cmd = f"""export TZ={timezone}; gunzip --stdout {data_file}  | sed -e 's/\\\\"/""/g' | sed -e "s/\\\\\\'/'/g" | clickhouse-client --use_client_time_zone 1 -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT CSV" -u{args.clickhouse_user} --password '{password}' -mn """
+            cmd = f"""export TZ={timezone}; gunzip --stdout {data_file}  | sed -e 's/\\\\"/""/g' | sed -e "s/\\\\\\'/'/g" | clickhouse-client --use_client_time_zone 1 -h {clickhouse_host} --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT CSV" -u{args.clickhouse_user} --password '{args.clickhouse_password}' -mn """
             logging.info(cmd)
             (rc, result) = run_quick_command(cmd)
             logging.debug(result)
@@ -447,7 +447,7 @@ def load_data_mysqlshell(args, timezone, schema_map, dry_run = False):
                 # double quote escape logic https://github.com/ClickHouse/ClickHouse/issues/10624
                 structure = columns.replace(
                     ",", " Nullable(String),")+" Nullable(String)"
-                cmd = f"""export TZ={timezone}; zstd -d --stdout {data_file}  | clickhouse-client --use_client_time_zone 1 --throw_if_no_data_to_insert=0  -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT TSV" -u{args.clickhouse_user} --password '{password}' -mn """
+                cmd = f"""export TZ={timezone}; zstd -d --stdout {data_file}  | clickhouse-client --use_client_time_zone 1 --throw_if_no_data_to_insert=0  -h {clickhouse_host} --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT TSV" -u{args.clickhouse_user} --password '{args.clickhouse_password}' -mn """
                 futures.append(executor.submit(execute_load, cmd))
 
         for future in concurrent.futures.as_completed(futures):
--- a/sink-connector/python/db_load/clickhouse_loader.py
+++ b/sink-connector/python/db_load/clickhouse_loader.py
@@ -204,8 +204,8 @@ def convert_to_clickhouse_table(user_name, table_name, source):
     src = re.sub(r'', '', src)
     # adding virtual columns ver and sign
     src = re.sub(r'\) ENGINE',
-                 '  `_sign` Int8 DEFAULT 1,\n  `_version` UInt64 DEFAULT 0\n) ENGINE', src)
-    src = re.sub(r'ENGINE=InnoDB[^;]*', 'ENGINE = ReplacingMergeTree(_version) ' +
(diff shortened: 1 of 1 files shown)


sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -202,8 +204,8 @@ def convert_to_clickhouse_table(user_name, table_name, source):
     src = re.sub(r'', '', src)
     # adding virtual columns ver and sign
     src = re.sub(r'\) ENGINE',
-                 '  `_sign` Int8 DEFAULT 1,\n  `_version` UInt64 DEFAULT 0\n) ENGINE', src)
```
subkanthi, line 205: we might want to keep this for older versions of Clickhouse without the `is_deleted` column. @aadant  WDYT?

aadant: Yes, actually I have a better version of this code that uses ANTLR for parsing and falls back to the regexp code.
I will send a PR soon. It should fix the comment issue !

aadant: see https://github.com/Altinity/clickhouse-sink-connector/pull/310 for a similar fix (except for the comment)
once we reconcile this PR and #310, we can merge this one https://github.com/Altinity/clickhouse-sink-connector/pull/309 that uses a proper DDL parser and falls back to the original regexp parser.

sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -194,6 +194,8 @@ def convert_to_clickhouse_table(user_name, table_name, source):
     # ClickHouse does not support constraints, indices, primary and unique keys
     src = re.sub(r'.*\bCONSTRAINT\b.*', '', src)
     src = re.sub(r'.*\bPRIMARY KEY\b.*\(.*', '', src)
+    # COMMENT is not supported
+    src = re.sub(r'\s*\bCOMMENT\b\s*\'(.*?)\'', '', src)
     # primary key on the column itself
```
aadant, line 199: feel free to add this to the code, it is a good addition to the regexp parser (there is a more elaborate ANTLR parser that kicks in first now), so it would only be called if you explicitly enable the regexp parser of fi there is an exception in the ANTL parser.

sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -202,8 +204,8 @@ def convert_to_clickhouse_table(user_name, table_name, source):
     src = re.sub(r'', '', src)
     # adding virtual columns ver and sign
     src = re.sub(r'\) ENGINE',
-                 '  `_sign` Int8 DEFAULT 1,\n  `_version` UInt64 DEFAULT 0\n) ENGINE', src)
-    src = re.sub(r'ENGINE=InnoDB[^;]*', 'ENGINE = ReplacingMergeTree(_version) ' +
+                 '  `is_deleted` UInt8 DEFAULT 0,\n  `_version` UInt64 DEFAULT 0\n) ENGINE', src)
+    src = re.sub(r'ENGINE=InnoDB[^;]*', 'ENGINE = ReplacingMergeTree(_version, is_deleted) ' +
                  partitioning_options + ' ORDER BY ('+primary_key+') SETTINGS '+settings, src)
```
aadant, line 209: This is merged already

sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -408,7 +410,7 @@ def load_data(args, timezone, schema_map, dry_run = False):
         for data_file in data_files:
             # double quote escape logic https://github.com/ClickHouse/ClickHouse/issues/10624
             structure = columns.replace(","," Nullable(String),")+" Nullable(String)"
-            cmd = f"""export TZ={timezone}; gunzip --stdout {data_file}  | sed -e 's/\\\\"/""/g' | sed -e "s/\\\\\\'/'/g" | clickhouse-client --use_client_time_zone 1 -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT CSV" -u{args.clickhouse_user} --password '{password}' -mn """
+            cmd = f"""export TZ={timezone}; gunzip --stdout {data_file}  | sed -e 's/\\\\"/""/g' | sed -e "s/\\\\\\'/'/g" | clickhouse-client --use_client_time_zone 1 -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT CSV" -u{args.clickhouse_user} --password '{args.clickhouse_password}' -mn """
             logging.info(cmd)
```
aadant, line 414: already merged

sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -445,7 +447,7 @@ def load_data_mysqlshell(args, timezone, schema_map, dry_run = False):
                 # double quote escape logic https://github.com/ClickHouse/ClickHouse/issues/10624
                 structure = columns.replace(
                     ",", " Nullable(String),")+" Nullable(String)"
-                cmd = f"""export TZ={timezone}; zstd -d --stdout {data_file}  | clickhouse-client --use_client_time_zone 1 --throw_if_no_data_to_insert=0  -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT TSV" -u{args.clickhouse_user} --password '{password}' -mn """
+                cmd = f"""export TZ={timezone}; zstd -d --stdout {data_file}  | clickhouse-client --use_client_time_zone 1 --throw_if_no_data_to_insert=0  -h {clickhouse_host} --secure --query="INSERT INTO {ch_schema}.{table_name}({columns})  SELECT {transformed_columns} FROM input('{structure}') FORMAT TSV" -u{args.clickhouse_user} --password '{args.clickhouse_password}' -mn """
                 futures.append(executor.submit(execute_load, cmd))
```
aadant, line 451: already merged

sink-connector/python/db_load/clickhouse_loader.py
```diff
@@ -495,7 +497,7 @@ def main():
     parser.add_argument('--dry_run', dest='dry_run',
                         action='store_true', default=False)
     parser.add_argument('--virtual_columns', help='virtual_columns',
-                        nargs='+', default=['`_sign`', '`_version`'])
+                        nargs='+', default=['`is_deleted`', '`_version`'])
```
aadant, line 500: it does not hurt to add it, let us keep `_sign`

aadant (changes requested):
Please see my comments

sink-connector/python/db_load/clickhouse_loader.py
AdamKatzDev: Right, I've noticed that in new parser you've decided to generate both `_sign` and `is_deleted` columns. It is a bit confusing but I guess you have reasons for this.

aadant: indeed, if you use `is_deleted` you do not need `_sign`, will fix. If the sink-connector sees the column, it will populate it. 
those are the kind of decisions should be made obvious with a replication.yml file.

aadant: https://github.com/Altinity/clickhouse-sink-connector/issues/313

AdamKatzDev:
@aadant
I've tried to reproduce the bug but could not. `COMMENT` is working just fine. Sorry for confusing you.
Probably my older fix had a faulty regex that removed the column with `COMMENT` altogether which helped to resolve the crashes in my case, the issue was in the regex parser somewhere else.

PR #339, opened 2023-10-24
Get last element as a table name from a topic name

IlyaTsoi:
It would be better if it didn't depend on the hard-coded <server>.<database>.<table> template

### DIFF
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/common/Utils.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/common/Utils.java
@@ -86,18 +86,14 @@ public class Utils {
      * @return Table Name
      */
     public static String getTableNameFromTopic(String topicName) {
-        String tableName = null;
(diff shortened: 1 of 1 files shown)


subkanthi:
DId you come across a bug for this, all the debezium connectors create topics in the following format.
```
<topicPrefix>.<schemaName>.<tableName>
```
Would be better to return a NULL table name instead of extracting a wrong table name and writing data to it.

IlyaTsoi:
In any case a table name is located at the end of the topic name. Isn't it? Current logic brakes when using topic.prefix in debezium config and we get `null` instead of table name. Of course `topic2table` mapping by regex will resolve this issue. My solution migth be used as workaround.

IlyaTsoi:
@subkanthi one more argument): when `getTableNameFromTopic()` returns `null` the table with this name will be created and the table will contain all columns from all source tables). Of course, replication will fail. I have not inspected why, but seen such behavior.

Issue #215, opened 2023-04-21
Wrong date format


sudmed:
Hello experts, I run the last Lightweight version (pg to ch) and on my dataset a date is in wrong format. 
In postgres table the date format is `timestamp with time zone` and it looks like `2023-04-21 21:43:11.965283+03`.
In clickhouse table the date format is `DateTime64(6)` and it looks like `2023-04-21 18:43:11.965283000`.
As you can see UTC+3 became UTC.

Is there a way to make the date in CH be the same as PG?

aadant:
Actually you can change the default timezone for ClickHouse server and it should solve the issue.
Reopen if it is not the case.

sudmed:
Hello, @aadant 
It is not the case, changing the default timezone for ClickHouse server didn't solve the issue. 
It seems that Lightweight Replicator doesn't understand `timestamp with time zone`. If I put `timestamp without time zone` in postgres, the Lightweight Replicator correctly streams it to CH.
You can reproduce the problem:

<details><summary>Create PG table</summary>
<p>

```
CREATE TABLE test
(
  id bigserial NOT NULL,
  consultation_id int8 NOT NULL,
  recomendation text NULL,
  date1 timestamp with time zone NOT NULL DEFAULT now(),
  date2 timestamp(0) with time zone NOT NULL DEFAULT now(),
  date3 timestamp with time zone NOT NULL DEFAULT '2023-04-21 21:43:11.965283+03',
  date4 timestamp(0) with time zone NOT NULL DEFAULT '2023-04-21 21:43:32+03',
  date5 timestamp without time zone NOT NULL DEFAULT now(),
  CONSTRAINT protocol_pkey PRIMARY KEY (id)
);
```

</p>
</details> 

<details><summary>Create CH table</summary>
<p>

```
CREATE TABLE test
(
    `id` Int64,
    `consultation_id` Int64,
    `recomendation` Nullable(String),
    `date1` DateTime64(9),
    `date2` DateTime,
    `date3` DateTime64(6),
    `date4` DateTime,
    `date5` DateTime64(6),
    `_sign` Int8,
    `_version` UInt64
)
ENGINE = ReplacingMergeTree(_version)
ORDER BY id;
``` 

</p>
</details> 


<details><summary>INSERT in PG</summary>
<p>

```
INSERT INTO protocol_test VALUES ('156', '56', 'ттест');
``` 

</p>
</details> 


<details><summary>result of Lightweight Replicator</summary>
[...]

PR #341, opened 2023-10-25
Use zoned time when parse string

IlyaTsoi:
I suggest to use ZonedDateTime instead of LocalDateTime. So we can parse a time_string with an offset marker e.g. +03:00, -0130, Z. Also we can handle an exception when a dateTime string has invalid year e.g. in MSSQL "0000-00-01 00:00:00". Zero year value is invalid in java.


IlyaTsoi:
Maybe it can solve #215?

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java
```diff
@@ -151,29 +152,53 @@ public static String convert(Object value) {
...
+                        setMinDate = true;
+                        break;
+                    }
                     // Continue
                 }
             }
-            if(parsingSuccesful == false) {
+
+            if (parsingSuccesful) {
+                if (setMinDate) {
```
subkanthi: If they are both true, why is there a need to have 2 variables?

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java
```diff
@@ -151,29 +152,53 @@ public static String convert(Object value) {
...
                 try {
                     DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatString);
-                    LocalDateTime zd = LocalDateTime.parse((String) value, formatter);
-                    result = zd.format(destFormatter);
-                    //result = removeTrailingZeros(result);
+                    parsedDT = ZonedDateTime.parse((String) value, formatter);
                     parsingSuccesful = true;
                     break;
                 } catch(Exception e) {
+                    if (e.getMessage().contains("Invalid value for YearOfEra")) {
```
subkanthi: Doesn't look like a good idea to check for exception using the message(Message could change in the future), isnt there a specific exception if parsing fails?

subkanthi:
Please add to this test case like this
```
    @Test
    public void testZonedTimestampConverter() {

        String formattedTimestamp = DebeziumConverter.ZonedTimestampConverter.convert("2021-12-31T19:01:00Z");
        Assert.assertTrue(formattedTimestamp.equalsIgnoreCase("2021-12-31 19:01:00.000000"));

        String formattedTimestampWMicroSeconds = DebeziumConverter.ZonedTimestampConverter.convert("2038-01-19T03:14:07.999999Z");
        Assert.assertTrue(formattedTimestampWMicroSeconds.equalsIgnoreCase("2038-01-19 03:14:07.999999"));

        String formattedTimestamp3 = DebeziumConverter.ZonedTimestampConverter.convert("2038-01-19T03:14:07.99Z");
        Assert.assertTrue(formattedTimestamp3.equalsIgnoreCase("2038-01-19 03:14:07.990000"));

        String formattedTimestamp4 = DebeziumConverter.ZonedTimestampConverter.convert("0000-00-01 00:00:00");
        System.out.println(formattedTimestamp4);
    }
````

File: sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java (hunk 1 above)
Comment (IlyaTsoi) [reply sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java]: When we get invalid year exception we can handle it by setting the setMinDate flag. So we don't get a date by parsing, but use the `setMinDate` flag to set it to minimum date. If this error occurs then we can safely assume that the date was not parsed due to zero year value

File: sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java (hunk 2 above)
Comment (IlyaTsoi) [reply sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java]: OK, i will check.

Comment (IlyaTsoi) [reply sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java]: I have rewritten a bit this part. It's possible to use `getCause()` method that inherited from `Throwable` to define if it's either a `DateTimeParseException` error or a `DateTimeException`. If `getCause()` returns `null` - it's a parsing error, otherwise not.

File: sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/common/Utils.java [hunk 3]
```diff
@@ -88,13 +88,9 @@ public static Map<String, String> parseTopicToTableMap(String input) throws Exce
...
 
-
-            // topic names is of the following format.
-            // hostname.dbName.tableName
+            // get last element after deviding topicName by "." as tableName
             String[] splitName = topicName.split("\\.");
-            if(splitName.length == 3) {
-                tableName = splitName[2];
-            }
+            tableName = splitName[splitName.length - 1];
```
subkanthi: please remove these changes for topic.prefix

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/converters/DebeziumConverter.java
```diff
@@ -151,29 +152,52 @@ public static String convert(Object value) {
...
             boolean parsingSuccesful = false;
+            ZonedDateTime parsedDT = null;
             for (String formatString : date_formats) {
                 try {
                     DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatString);
-                    LocalDateTime zd = LocalDateTime.parse((String) value, formatter);
-                    result = zd.format(destFormatter);
-                    //result = removeTrailingZeros(result);
+                    parsedDT = ZonedDateTime.parse((String) value, formatter.withZone(ZoneId.of("UTC")));
                     parsingSuccesful = true;
```
subkanthi: nit, parsingSuccessful

subkanthi:
This was implemented as part of other PR's , closing this for now.

PR #343, opened 2023-10-25
Fix the getVersionColumnForReplacingMergeTree() to support replicated tables properly

IlyaTsoi:
1. In the current logic the getVersionColumnForReplacingMergeTree() returns only version_column from a replicated table definition (here `versionColumn = parameterArray[2].trim();`), but a is_deleted column migth be there too.

2. It's unnecessary to have separate if clauses for replicated and nonreplicated tables, we can union them.


subkanthi:
Please check `DBMetadataTest` and add your new SQL that you are trying to fix.

Issue #349, opened 2023-10-26
Specify timezone for DateTime / DateTime64 


aadant:
Specify the default timezone for the CH server :

```
clickhouse.datetime.timezone=UTC
```

This setting would control the timezone for DateTime* data types : 
```
`ts` DateTime64(6, 'UTC')
```

aadant:
This is related to the `source` setting, ideally both are the same for usual migration / replication :
```
database.connectionTimeZone: "America/Chicago"
```

IlyaTsoi:
@aadant,  I'm trying to solve this. I think that would be better if it didn't solve on a column data type level, but on `INSERT` query level. We can use `UTC` by default within sink-connector and use `toDateTime*()` within `INSERT` query with indicating time zone:
```sql
INSERT INTO table(`date_column`) VALUES (toDateTime64('2023-04-29 06:36:28.46277', 6, 'UTC'))
```
Or always use `integer` for `DateTime` types within `INSERT` query. This is from Clickhouse docs:
```text
When inserting datetime as an integer, it is treated as an appropriately scaled Unix Timestamp (UTC)
```
So, it would be nice if `ZonedTimestampConverter()` returned an `integer` instead of a `string`.

aadant:
@IlyaTsoi  sure if `clickhouse.server.default_timezone` is not specified, it falls back to the server time zone. However, in order to :

- emulate the DateTime behavior of MySQL (insensitive to TZ change)
- source and target servers timezone difference
- source table representing UTC DateTime replicating to a non UTC server

we need this override.

Currently the sink-connector casts all DateTimes to UTC and it is influenced by `database.connectionTimeZone`.

UTC -> UTC works
TZ -> same TZ works if `database.connectionTimeZone=TZ`(@subkanthi tested that, I need to double check)
 
Ideally we need to override at the column level (overrides are not implemented yet, except at the global level)
see https://github.com/Altinity/clickhouse-sink-connector/issues/323

aadant:
Actually the difficulty is that DateTime In MySQL is like String representing Date and Time, while DateTime* in CH behave like MySQL Timestamp. In addition and unlike MySQL, CH can stick the TZ at the column level. 

https://dev.mysql.com/doc/refman/8.0/en/datetime.html

IlyaTsoi:
> CH can stick the TZ at the column level.

[...]

PR #374, opened 2023-11-13
Param source timezone (review)

aadant:
Looks good to me apart from a clash with #366.  I managed to replicate to UTC with this PR.


sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java
```diff
@@ -141,6 +147,29 @@ public ResultSet executeQueryWithResultSet(String sql) throws SQLException {
...
+        while (matcher.find()) {
+            int start = matcher.start();
+            int end = matcher.end();
+            result.append(typeName.substring(cursor, start));
+            cursor = end;
+            String occurrence = typeName.substring(start, end);
+            if (occurrence.contains("DateTime64")) {
+                String[] params = StringUtils.substringBetween(occurrence, "(", ")").split(",");
+                String precision = params[0].trim();
+                result.append(String.format("DateTime64(%s,\\'%s\\')", precision, timeZone));
```
aadant, line 164: escape clashing with another escape

aadant: in #366  
```
String dataType = ClickHouseUtils.escape(columnNameToDataTypeMap.get(entry.getKey()), '\'');
```

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java
```diff
@@ -141,6 +147,29 @@ public ResultSet executeQueryWithResultSet(String sql) throws SQLException {
 
     }
 
+    public String addTimeZoneToColumnDefinition(String typeName, String timeZone) {
```
aadant, line 150: not sure it is a good idea to use a regexp here as the target DateTime64 may already have a TZ !

example  DateTime64(6,'UTC') becomes
  DateTime64(6,'America/Chicago')

aadant: I wonder if it works if you use a String instead, CH will do the conversion

Exception I got with the maximum boundaries
```
`db_to` DateTime64(6,\'America/Chicago\'),`_version` UInt64,`is_deleted` UInt8')
118741 2023-11-13 02:53:56.165 [pool-1-thread-9] ERROR com.altinity.clickhouse.sink.connector.db.DbWriter  - ******* ERROR inserting Batch *****************
java.lang.IllegalArgumentException: DateTime(9904571999) should be between -1420070400 and 9904550399 inclusive of both values
	at com.clickhouse.data.ClickHouseChecker.newException(ClickHouseChecker.java:17)
	at com.clickhouse.data.ClickHouseChecker.between(ClickHouseChecker.java:109)
	at com.clickhouse.data.format.BinaryDataProcessor$DateTime64SerDe.serialize(BinaryDataProcessor.java:292)
	at com.clickhouse.data.ClickHouseDataProcessor.write(ClickHouseDataProcessor.java:534)
	at com.clickhouse.jdbc.internal.InputBasedPreparedStatement.addBatch(InputBasedPreparedStatement.java:345)
	at com.altinity.clickhouse.sink.connector.db.DbWriter.addToPreparedStatementBatch(DbWriter.java:479)
	at com.altinity.clickhouse.sink.connector.executor.ClickHouseBatchRunnable.flushRecordsToClickHouse(ClickHouseBatchRunnable.java:199)
	at com.altinity.clickhouse.sink.connector.executor.ClickHouseBatchRunnable.processRecordsByTopic(ClickHouseBatchRunnable.java:169)
	at com.altinity.clickhouse.sink.connector.executor.ClickHouseBatchRunnable.run(ClickHouseBatchRunnable.java:101)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

aadant (commented):
This implementation works for UTC but does not for other timezone if the maximum range is reached.

the solution is to let CH convert according the target data type.

https://github.com/Altinity/clickhouse-sink-connector/blob/418d4ec8f2b0d326febe14b78e3a9eee86afeb86/sink-connector/python/db_load/clickhouse_loader.py#L499

I wonder if this works for say Chicago to Chicago instead of UTC to UTC.
The source.datetime.timezone should not be defaulted to UTC (if not specified) but the debezium source timezone.

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/ClickHouseSinkConnectorConfig.java
```diff
@@ -263,6 +263,17 @@ static ConfigDef newConfigDef() {
                     1,
                     ConfigDef.Width.NONE,
                     ClickHouseSinkConnectorConfigVariables.REPLACING_MERGE_TREE_DELETE_COLUMN.toString())
+                .define(
+                    ClickHouseSinkConnectorConfigVariables.SOURCE_DATETIME_TIMEZONE.toString(),
+                    Type.STRING,
+                    "UTC",
```
aadant, line 269: not sure it is a good idea. leave it `null` by default and initialize with source timezone from Debezium.
Add to the default config.yaml (commented)

sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java
IlyaTsoi: This is quite logical. We can take into account source.timezone for `checkIfDateExceedsSupportedRange()` methods.

aadant: yes that works. @subkanthi FYI

IlyaTsoi: > not sure it is a good idea to use a regexp here as the target DateTime64 may already have a TZ !
> 
> example DateTime64(6,'UTC') becomes DateTime64(6,'America/Chicago')

Could you explain, please? I don't see any problem here (except possible slowness)). I specifically override it within `input()` function to make Clickhouse use defined `source.timezone` to parse incoming data instead of column metadata.

aadant: It is the wrong place to change the schema. The sink-connector should create the schema with the correct timezone (target timezone).

aadant: in the sink-connector, there are 2 places we fix the schema 
- for the initial snapshot via custom tools like https://github.com/Altinity/clickhouse-sink-connector/pull/375/files
- via debezium, the sink-connector needs to add this `clickhouse.datetime.timezone` parameter. In your case, it would be the same as the source. default should be not specified and defaults to `database.connectionTimeZone` (debezium)
`source.datetime.timezone` should also default to `database.connectionTimeZone`

what do you think ?

aadant: in terms of timing, let us wait for @subkanthi to merge all pending PRs to develop, we will then address this one.

aadant: @subkanthi  let us implement this too https://github.com/Altinity/clickhouse-sink-connector/issues/349

subkanthi:
Thanks @IlyaTsoi for your contribution.

PR #379, opened 2023-11-17
Fix datetime limits

### LINKED ISSUE #373: Add logic to limit DateTime max
```debezium-embedded_1  | java.lang.IllegalArgumentException: DateTime(-1420092000) should be between -1420070400 and 9904550399 inclusive of both values```

Currently only `DateTime64` limits are checked in the sink connector, if the `DateTime` limits exceeds UInt64, `clickhouse-jdbc` throws an error.

Solution: Limit the value before inserting into CH.

subkanthi:
- Added logic to retrieve clickhouse server timezone.
- Added integration tests to run data_types.sql with Ch server timezone set to `America/Chicago`

Closes: #373


aadant (approved):
LGTM

PR #404, opened 2023-12-07
Rework docker-compose-mysql.yml example used in the quikcstart guide

Selfeer:
Rework the docker-compose-mysql.yml file used for a quickstart guide.

**Updates:**

- Moved services used inside a main compose file into separate `service` files
- For the Grafana, instead of building from a dockerfile that copies config files, we have a grafana service that instead mounts config files as volumes and uses `grafana/grafana:latest` image.


sink-connector-lightweight/docker/clickhouse-sink-connector-lt-service.yml
```diff
@@ -0,0 +1,15 @@
+version: "3.4"
+
+services:
+  clickhouse-sink-connector-lt:
+    image: ${CLICKHOUSE_SINK_CONNECTOR_IMAGE}
```
subkanthi: can we update the quickstart.md with this new environment variable.

Issue #459, opened 2024-02-08
Values of the generated columns in MySQL are not transfered to the ClickHouse table


Selfeer:
If we create a table in MySQL that has a generated column it is created in ClickHouse but if in MySQL we insert the data into that table it is not transferred to ClickHouse table.

For example,
```sql
CREATE TABLE test_table (
  id INT AUTO_INCREMENT PRIMARY KEY NOT NULL, 
  first_name VARCHAR(50) NOT NULL,
  last_name VARCHAR(50) NOT NULL,
  fullname varchar(101) GENERATED ALWAYS AS (CONCAT(first_name,' ',last_name)),
  email VARCHAR(100) NOT NULL);

INSERT INTO test_table (id, first_name, last_name, email) VALUES (1, "Test", "Name", "test@gmail.com");

SELECT * FROM test_table;

+----+------------+---------------+---------------------+------------------------------+
| id | first_name | last_name |    fullname        |        email                     |
+----+------------+---------------+---------------------+------------------------------+
|  1 | Test      | Name           |   Test Name     |    test@gmail.com        |
+----+------------+---------------+---------------------+------------------------------+
1 row in set (0.00 sec)
```

If after inserting the value we check the table in ClickHouse we see that it's empty
```sql
SELECT *
FROM test_table;

Query id: e8c78e7b-dea9-49aa-a735-1c4c59f6990b

Ok.

0 rows in set. Elapsed: 0.003 sec.

SHOW CREATE TABLE test_table;

CREATE TABLE test.test_table
(
    `id` Int32,
    `first_name` String,
    `last_name` String,
    `fullname` Nullable(String) ALIAS concat(first_name, ' ', last_name),
    `email` String,
    `_version` UInt64,
    `is_deleted` UInt8
)
ENGINE = ReplacingMergeTree(_version, is_deleted)
ORDER BY id
SETTINGS index_granularity = 8192
```

**The build used:** `altinityinfra/clickhouse-sink-connector:443-14a14ba5c18fb3bc9f57c48d9a5079d4d6fe15e8-lt`
**ClickHouse Version**: `23.11.1.2711`
**MySQL Version**: `8.2.0`

The error in logs
```
[...]

PR #515, opened 2024-03-26
Various fixes to the MySQL python commands

aadant:
mysql_dumper.py 

- support for partitioned tables only, without the `--partitioned_table_only` flag, tables and partitioned matching the regexps are dumped
- support for very large dump definitions with a lot of partitions
- all tables are selected by default (`--include_tables_regex .`) 

mysql_parser.py

- python ANTLR translation skips the collation that can be found in generated column definitions.
- support for generated columns see https://github.com/Altinity/clickhouse-sink-connector/issues/459

sink-connector/python/db_compare/clickhouse_table_checksum.py
- added a `max_datetime_value` to even out sub-second discrepancies in the max datetimes (overflows)

### DIFF
--- a/sink-connector/python/db_compare/clickhouse_table_checksum.py
+++ b/sink-connector/python/db_compare/clickhouse_table_checksum.py
@@ -1,5 +1,5 @@
 """
-# -- ============================================================================
+#r/ -- ============================================================================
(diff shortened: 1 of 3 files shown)


minguyen9988 (approved):
lgtm

PR #582, opened 2024-05-09
feat: added SharedReplacingMergeTree support

vpol:
Tables created by clickhouse.cloud have SharedReplacingMergeTree engine, which currently is not handled correctly.


subkanthi:
> Tables created by clickhouse.cloud have SharedReplacingMergeTree engine, which currently is not handled correctly.

Hi @vpol , thanks for your contribution, we have added it to `2.2.0` release, it will be great to have integration tests for this similar to the one for ReplicatedReplacingMergeTree - https://github.com/Altinity/clickhouse-sink-connector/blob/develop/sink-connector-lightweight/src/test/java/com/altinity/clickhouse/debezium/embedded/ReplicatedRMTIT.java

vpol:
Hi. Thank you for reply.

After some quick research I discovered that SharedMergeTree (and SharedReplacingMergeTree) is only available in clickhouse.cloud, not sure how I can write tests in this case.

From: 

https://clickhouse.com/docs/en/cloud/reference/shared-merge-tree

> SharedMergeTree Table Engine

> * Available exclusively in ClickHouse Cloud (and first party partner cloud services)

> The SharedMergeTree table engine family is a cloud-native replacement of the ReplicatedMergeTree engines that is optimized to work on top of shared storage (e.g. Amazon S3, Google Cloud Storage, MinIO, Azure Blob Storage). There is a SharedMergeTree analog for every specific MergeTree engine type, i.e. ReplacingSharedMergeTree replaces ReplacingReplicatedMergeTree.

> The SharedMergeTree table engine family powers ClickHouse Cloud. For an end-user, nothing needs to be changed to start using the SharedMergeTree engine family instead of the ReplicatedMergeTree based engines. It provides the following additional benefits:

From what I understand it works exactly the same way as ReplicatedReplacingMergeTree, and was not supported purely because it has a different name and is not present in enum's.

aadant:
@vpol out of curiosity, I thought ClickHouse cloud would automatically turn the MergeTree tables of any kind into their Shared version.

vpol:
it does, but  it looks like it's turning it into a different "proprietary" table engine, with different name. name that is not correctly handled by the current code.

PR #594, opened 2024-05-15
Minor improvements to python scripts

aadant:
- `clickhouse_table_count.py` : use `do_not_merge_across_partitions_select_final`
- `mysql_table_count.py` : parallel counts per table
- `clickhouse_loader.py` : timestamp converted to DateTime64 with timezone have single quotes that need to be escaped.


minguyen9988 (approved):
LGTM

subkanthi (approved):

PR #685, opened 2024-07-12
Removed count assertions in postgres tests

subkanthi:


subkanthi:
closings, tests fixed in another PR.

PR #837, opened 2024-09-22
feat: Allow specifying the configmap name/contents

jaitaiwan:
This change moves the contents of the configmap to the values.yaml file and allows changing the configuration options directly or zero them out and specifying an external configmap that the deployment should use.


jaitaiwan:
Closing out in favour of merging to 2.4.0

Issue #976, opened 2025-01-23
ANTLR parsing error - Debezium


subkanthi:
https://issues.redhat.com/browse/DBZ-8600

```

CREATE DEFINER=`bcadmin`@`%` PROCEDURE `sp_next_available_otc_instance_strategy_id`()
begin
  select min(st.value) as strategy_id
  from SEQUENCE_TABLE(100000) st
  join btc_quant.stratId_ranges r on r.stratId0 <= st.value and st.value < r.stratId1
  left join otc_instance i on i.strategy_id=st.value
  where r.category = 'otc' and now() between validFromDate and coalesce(validToDate, now())
  and i.strategy_id is null;
end

```

 

```

no viable alternative at input '(100000'
    at io.debezium.antlr.ParsingErrorListener.syntaxError(ParsingErrorListener.java:43) ~[app.jar:?]
    at org.antlr.v4.runtime.ProxyErrorListener.syntaxError(ProxyErrorListener.java:41) ~[app.jar:?]
    at org.antlr.v4.runtime.Parser.notifyErrorListeners(Parser.java:543) ~[app.jar:?]
    at org.antlr.v4.runtime.DefaultErrorStrategy.reportNoViableAlternative(DefaultErrorStrategy.java:310) ~[app.jar:?]
    at org.antlr.v4.runtime.DefaultErrorStrategy.reportError(DefaultErrorStrategy.java:136) ~[app.jar:?]
    at io.debezium.ddl.parser.mysql.generated.MySqlParser.sqlStatements(MySqlParser.java:1280) ~[app.jar:?]
    at io.debezium.ddl.parser.mysql.generated.MySqlParser.root(MySqlParser.java:996) ~[app.jar:?]
    at io.debezium.connector.mysql.antlr.MySqlAntlrDdlParser.parseTree(MySqlAntlrDdlParser.java:82) ~[app.jar:?]
    at io.debezium.connector.mysql.antlr.MySqlAntlrDdlParser.parseTree(MySqlAntlrDdlParser.java:49) ~[app.jar:?]
    at io.debezium.antlr.AntlrDdlParser.parse(AntlrDdlParser.java:76) ~[app.jar:?]
    at io.debezium.connector.binlog.BinlogDatabaseSchema.parseDdl(BinlogDatabaseSchema.java:311) ~[app.jar:?]
[...]

PR #1001, opened 2025-03-02
Update debezium version to 3.1.0.AlphaFinal

subkanthi:
closes: #976


subkanthi:
image

`snapshot` was changed from boolean to String in Debezium 3.1.0.

PR #1041, opened 2025-04-11
fix: clickhouse_loader port and secure options

philippevezina:
`clickhouse_port` was not passed to some `clickhouse-client` commands and the `clickhouse_secure` was not available.


aadant (approved):
LGTM, thanks for the contribution !

Issue #1046, opened 2025-04-14
NullPointerException when dbSource is null in initiateNewConnectionIfClosed method


chimeyrock999:
### Problem:
In the `initiateNewConnectionIfClosed` method, there is a risk of a `NullPointerException` being thrown when the `dbSource` is `null`. This happens if the database name does not exist in the `instance` map, leading to an unhandled `null` value being used when attempting to get a connection from the data source.

### Steps to Reproduce:
1. Call the `initiateNewConnectionIfClosed` method with a database name that is not present in the `instance` map.
2. The method will attempt to retrieve a connection from a `null` `HikariDataSource`, causing a `NullPointerException`.

### Expected Behavior:
The method should properly handle the case when `dbSource` is `null` by throwing a meaningful exception or logging the error, preventing a `NullPointerException` from being thrown.

### Proposed Solution:
- Check if `dbSource` is `null` before attempting to get a connection.
- If `dbSource` is `null`, throw an `SQLException` with a meaningful message indicating the issue.

### Example Code Fix:
```java
public static Connection initiateNewConnectionIfClosed(String databaseName) throws SQLException {
    if (disabled) {
        return null;
    }
    
    HikariDataSource dbSource = instance.get(databaseName);
    
    if (dbSource == null) {
        throw new SQLException("Database source is null for database: " + databaseName);
    }
    
    HikariDbSource.printConnectionInfo();
    return dbSource.getConnection();
}
```

### Logs for this case
```java
15:14:11.387 [Sink Connector thread-pool-8] ERROR com.altinity.clickhouse.sink.connector.db.batch.GroupInsertQueryWithBatchRecords - **** ERROR ALTER TABLE: lineitem
java.lang.NullPointerException: Cannot invoke "com.zaxxer.hikari.HikariDataSource.getConnection()" because "dbSource" is null
	at com.altinity.clickhouse.sink.connector.db.HikariDbSource.initiateNewConnectionIfClosed(HikariDbSource.java:46) ~[clickhouse-kafka-sink-connector-0.0.9.jar:?]
	at com.altinity.clickhouse.sink.connector.db.DBMetadata.getColumnsDataTypesForTable(DBMetadata.java:340) ~[clickhouse-kafka-sink-connector-0.0.9.jar:?]
[...]

PR #1048, opened 2025-04-15
[1046] Fix NullPointerException and Refactor Connection Pool Initialization Logic

chimeyrock999:
## Description: 
1. Throw `Exception` when `dbSource` is `null`.
2. Refactor `HikariDataSource` initialization logic in `getInstance` method. Ensure that closed `HikariDataSource` instances will be removed to prevent memory leak. 
2. Change `HikariDbSource` from `HashMap` to `ConcurrentHashMap` for more thread safety.
## Benefits:
1. Avoids potential race conditions and redundant synchronization blocks.
2. Ensures correct connection reuse and recreation logic.
3. Clean, thread-safe code using idiomatic ConcurrentHashMap pattern.


subkanthi:
closes: #1046

Issue #1071, opened 2025-05-30
Sink Connector Kafka Stops Replicating With `connection.pool.disable: false`


Selfeer:
With the release `2.6.0` when we have set the `connection.pool.disable: false` the replication just stops at some point.

Here are the steps after which we have hit that issue:

```
                 3ms              [bash-tools] bash# cat <<EOF | curl --request POST --url "http://clickhouse-sink-connector-kafka:8083/connectors" --header 'Content-Type: application/json' --data @-
                 4ms              [bash-tools] >           {
                 5ms              [bash-tools] >   "name": "sink-connector",
                 5ms              [bash-tools] >   "config": {
                 7ms              [bash-tools] >     "connector.class": "com.altinity.clickhouse.sink.connector.ClickHouseSinkConnector",
                 8ms              [bash-tools] >     "tasks.max": "100",
                 9ms              [bash-tools] >     "topics": "SERVER5432.test.autocreate_aef0edcc_3d50_11f0_8b45_960004562c63",
                10ms              [bash-tools] >     "clickhouse.topic2table.map": "",
                11ms              [bash-tools] >     "clickhouse.server.url": "clickhouse",
                12ms              [bash-tools] >     "clickhouse.server.user": "root",
                13ms              [bash-tools] >     "clickhouse.server.password": "root",
                14ms              [bash-tools] >     "clickhouse.server.database": "test",
                15ms              [bash-tools] >     "clickhouse.server.port": 8123,
                16ms              [bash-tools] >     "clickhouse.table.name": "users",
                17ms              [bash-tools] >     "key.converter": "io.confluent.connect.avro.AvroConverter",
                18ms              [bash-tools] >     "value.converter": "io.confluent.connect.avro.AvroConverter",
                20ms              [bash-tools] >     "key.converter.schema.registry.url": "http://schemaregistry:8081",
                22ms              [bash-tools] >     "value.converter.schema.registry.url": "http://schemaregistry:8081",
                29ms              [bash-tools] >     "store.kafka.metadata": true,
                30ms              [bash-tools] >     "topic.creation.default.partitions": 6,
                31ms              [bash-tools] >     "store.raw.data": false,
                32ms              [bash-tools] >     "store.raw.data.column": "raw_data",
                33ms              [bash-tools] >     "metrics.enable": true,
                34ms              [bash-tools] >     "metrics.port": 8084,
                35ms              [bash-tools] >     "buffer.flush.time.ms": 500,
[...]

PR #1073, opened 2025-05-31
Replaced redpanda images.

subkanthi:
closes: #1071


Selfeer:
I don't think that fixes the issue, the testflows tests already use that image: https://github.com/Altinity/clickhouse-sink-connector/blob/459cd718461c7a60085413f906a315b7c8d0a517/sink-connector/tests/integration/env/kafka-service.yml#L7

PR #1139, opened 2025-09-16
Various improvements to checksums

aadant:
This PR introduced a new features and many bug fixes  : 

`top_level_table_checksum.py` 

- support for multiple databases

```
source:
    mysql:
        host: mysql1
        databases:
          - test
        ignored_columns:
            # https://github.com/Altinity/clickhouse-sink-connector/issues/1136
            - test.bit_table.bit1_column
           # https://github.com/Altinity/clickhouse-sink-connector/issues/1137
            - test.json_table.json_column

replicas:
    - clickhouse:
        host: clickhouse1
    - clickhouse:
        host: clickhouse2

```
- `--lock_tables_on_source` locks tables on the source (it should be a replica not to block writes but only replication)
- `--sleep_after_lock` sleeps after lock to wait for data to replicate

`mysql_table_checksum.py`

- escapes table names
- ability to exclude columns
- attempts to format JSON columns (compact format)

`clickhouse_table_checksum.py`

- fix checksums with NULL values in the first column
- trim the trailing zeroes in DateTimes similar to MySQL checksums
- support for `is_deleted` in the source table (previously it was ignored, leading to different checksums

### DIFF
--- a/sink-connector/python/db/mysql.py
+++ b/sink-connector/python/db/mysql.py
@@ -24,7 +24,6 @@ def get_mysql_connection(mysql_host, mysql_user, mysql_passwd, mysql_port, mysql
     conn = engine.connect()
     return conn
 
-
 def get_tables_from_regex_sql(conn, no_wc,  mysql_database, include_tables_regex, exclude_tables_regex=None, non_partitioned_tables_only=False, include_partitions_regex=None):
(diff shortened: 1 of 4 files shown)


minguyen9988 (approved):
LGTM

PR #1277, opened 2026-04-15
chore(clickhouse_jdbc): update maven and clickhouse jdbc

botriki:
1. sink-connector/pom.xml & sink-connector-lightweight/pom.xml

- Upgraded clickhouse-jdbc from 0.6.0/0.6.5 → 0.9.8. The old shaded/http classifier no longer exists in 0.9.x; sink-connector-lightweight now uses the all-dependencies classifier instead.
- Removed the explicit clickhouse-jdbc dependency from sink-connector/pom.xml (it was declared with the obsolete http classifier and is now pulled in transitively).
- Upgraded Lombok from 1.18.22 → 1.18.34 in both modules. Fixes ExceptionInInitializerError on TypeTag that occurs with JDK 17.0.12+.
- Added explicit maven-compiler-plugin 3.15.0 configuration with annotationProcessorPaths pointing to Lombok in both modules. Starting from plugin version 3.12, annotation processors are no longer auto-discovered from the compile classpath, so without this block Lombok generates no code and the build fails with cannot find symbol.

2. sink-connector-lightweight/dependency-reduced-pom.xml

- Mirrors the same maven-compiler-plugin and Lombok changes described above (this file is auto-generated by maven-shade-plugin from the lightweight module's POM).
- Removed the commons-compress exclusion that was no longer needed.

3. SinkConnectorDataSource.java

- Changed getConnection() return type from ClickHouseConnection to java.sql.Connection to comply with the clickhouse-jdbc v2 API (the concrete type is no longer part of the public interface).
- Added IGNORE_UNSUPPORTED_KEY = "jdbc_ignore_unsupported_values" constant and two private helpers (ensureIgnoreUnsupported, withIgnoreUnsupported) that automatically append jdbc_ignore_unsupported_values=true to every URL / Properties object passed to the constructor. This prevents SQLFeatureNotSupportedException when standard JDBC callers invoke setAutoCommit(false), commit(), or rollback() — operations that ClickHouse does not support.

4. DebeziumChangeEventCapture.java

- Added calls to ensureIgnoreUnsupportedValuesParam(props, "offset.storage.jdbc.url") and ensureIgnoreUnsupportedValuesParam(props, "schema.history.internal.jdbc.url") right after the driver class is loaded. This patches Debezium's internal JDBC storage URLs with the same jdbc_ignore_unsupported_values=true flag so that Debezium's RetriableConnection wrapper no longer throws when it calls setAutoCommit(false) on a ClickHouse connection.
- Added the ensureIgnoreUnsupportedValuesParam static helper method and the matching IGNORE_UNSUPPORTED_KEY constant.


subkanthi:
Thanks for your contribution @botriki

Cong-Ye:
I also try to upgrade it, but got error:
image
There's a workaround, but better to change jdbc related code to fix it:
image

PR #1278, opened 2026-04-18
Clickhouse jdbc 2 10

subkanthi:
created from https://github.com/Altinity/clickhouse-sink-connector/pull/1277 so that the actions(CI/CD) can be tested.


minguyen9988:
## Full test + edge-case verification, and dual JDBC driver support

Branch updated to `73cc2af` (tree verified byte-identical to the locally tested working copy). Base `2.10.0` merged in — the branch had been 14 commits behind.

Everything below was verified against a **live ClickHouse 24.8.8** server with **clickhouse-jdbc 0.9.8**, on **both** driver generations. Every fix has a test that fails when the fix is reverted.

### 1. Dual JDBC driver support (old + new, one build)

`clickhouse-jdbc` 0.7+ ships **both** driver implementations in a single jar. This is now exposed as a connector parameter, so no separate fat jar or recompile is needed:

```
clickhouse.jdbc.v1 = false   (default)  -> V2 driver, com.clickhouse.jdbc.ConnectionImpl
clickhouse.jdbc.v1 = true               -> legacy V1, com.clickhouse.jdbc.internal.ClickHouseConnectionImpl
```

Plumbed through `SinkConnectorDataSource`, `BaseDbWriter.createConnection`, HikariCP, and the Debezium offset-storage / schema-history URLs. The V2-only `jdbc_ignore_unsupported_values` flag is deliberately not sent to V1 (which warns on unknown URL params).

Covered by 17 tests including live round-trips through **each** driver and **cross-driver readback** (V2 writes / V1 reads, and the reverse), 100 rows checksum-verified in both directions.

### 2. Three real defects found by running the suite on the V2 driver

#### 2a. `DBMetadata.getColumnsDataTypesForTable` — underscore wildcard column corruption

The method used `DatabaseMetaData.getColumns(null, database, tableName, null)`. `getColumns()` treats its schema/table arguments as **JDBC LIKE patterns**, where `_` matches *any* single character. So a table named `under_score_t` also matched `underXscore_t`, and the columns of **both** tables were merged into one column map.

A wrong column map produces a wrong INSERT column list — silent data corruption. Underscores are ubiquitous in replicated MySQL table names.

Live evidence:

```
V2 getColumns(null,'pr_db','under_score_t',null) = [decoy_a, decoy_b, id, v]
V1 getColumns(null,'pr_db','under_score_t',null) = [decoy_a, decoy_b, id, v]
system.columns(pr_db, under_score_t)             = [id, v]
```

Note this reproduces on **both** driver generations — it is a latent defect in the existing code, not a V1→V2 migration regression. Replaced with an exact-equality `system.columns` query, matching how the sibling metadata methods in the same class already read column metadata.

New regression test `DBMetadataTest.testGetColumnsDataTypesForTableWithUnderscoreDoesNotMatchSiblingTable`. **Negative control:** with the fix reverted it fails `expected:<2> but was:<4>`.

#### 2b. `ClickHouseDataTypeMapper` — geo (Point/Polygon) binding

The two drivers accept **mutually exclusive** representations; neither single form works on both. Measured live:

```
V1 setObject(ClickHouseGeoPointValue) = OK
V1 setString("(1.0,2.0)")             = FAIL UnsupportedOperationException
                                        "Converting [String] to [Point] is not supported"
V2 setObject(ClickHouseGeoPointValue) = FAIL Code: 47 UNKNOWN_IDENTIFIER
                                        "Missing columns: 'ClickHouseGeoPointValue'
                                         ... 'ClickHouseGeoPointValue[(1., 2.)]'"
V2 setString("(1.0,2.0)")             = OK
```

Polygon behaves identically. Added `setGeoValue()`, which picks the representation the active driver understands. Driver detection is `ps.isWrapperFor(ClickHousePreparedStatement.class)` — verified `true` only on the V1 statement (`InputBasedPreparedStatement`) and `false` on the V2 statement (`PreparedStatementImpl`).

#### 2c. Driver-agnostic reads in two ITs

- `CreateTableDataTypesIT`: a `Point` reads back as a `String` under V1 and a `double[]` under V2 — both are now normalized to `(x,y)` before asserting.
- `CreateTableDataTypesTimeZoneIT`: the V2 driver correctly refuses `getTimestamp()` on a `Date32` column; switched to `getDate()`, which the assertions on the following lines already used.

### Test results

```
sink-connector module (live containers, podman):
  115 tests, 0 failures, 2 skipped     (114 before; +1 new regression test)

Negative controls (fix reverted, test kept):
  underscore test          -> FAIL  expected:<2> but was:<4>
  geo setObject/setString  -> failures reproduced exactly as in the matrix above

Pre-existing, NOT caused by this PR:
  CreateTableDataTypesIT fails identically WITH and WITHOUT these changes
  (NPE at line 87, missing `datatypes` fixture columns). Confirmed by running
  the same test against the unmodified DBMetadata on this branch.
```

The new regression test sets `connection.pool.disable=true` so it cannot disturb the Hikari pools shared by the other tests in its class.

### Backward / forward compatibility

- Default behavior is unchanged (V2 driver), so existing deployments are unaffected.
- Setting `clickhouse.jdbc.v1=true` restores full legacy behavior from the same artifact.
- Cross-driver readback is verified in both directions, so data written by either driver is readable by the other.

minguyen9988:
## CI follow-up: `checksum-tests-lightweight` failure root-caused and fixed

Branch updated to `aac31d9`.

The previous head (`73cc2af`) failed `checksum-tests-lightweight`. I checked whether that was pre-existing: across the last 37 CI runs of that job it passed **36 times** and failed only on this branch — so it was **not** a flake and not pre-existing. It is a real bug that this PR's `clickhouse-jdbc` **0.6.0 → 0.9.8** upgrade makes reachable.

### What actually happened

The connector never created the 8 destination tables. It logged **32,690** repetitions of `Database altinity_sink_connector does not exist` across the full 10-minute window, and the checksum fixture timed out waiting for replication parity.

The chain:

1. `setupDebeziumEventCapture()` creates the system-database connection **exactly once** and reuses it for the Debezium storage-database creation and the version lookup.
2. `BaseDbWriter.createConnection()` logs and swallows a connect failure, returning `null`.
3. The compose stacks set `connection.pool.disable=true`, so there is **no pool** to obtain a replacement connection from later. `DBMetadata.executeSystemQuery()` correctly refuses to burn its retry budget in that case (`"no connection pool can provide one, giving up"`).
4. So a single `null` at startup wedges the process for its entire lifetime.

### Why the JDBC upgrade exposes it

compose gates the connector on the ClickHouse healthcheck, but the healthcheck can pass moments before the HTTP port is actually serving. The two driver generations behave differently in that window — measured live with clickhouse-jdbc 0.9.8 against a port with nothing listening:

```
V2 driver -> connection object returned  (connect is deferred/lazy)
V1 driver -> SQLException "Connect to http://host:port failed: Connection refused"
             => BaseDbWriter.createConnection() returns null
```

On 0.6.0 the startup happened to survive the window; on 0.9.8 it can return the hard failure. Whether a cold start worked was effectively depending on which driver generation defers the connect — which is not something startup should rely on either way.

### Fix

Obtain the initial system-database connection through a retry loop — 30 attempts × 2 s = 60 s, comfortably longer than the compose healthcheck's 30 s `start_period` — instead of accepting the first `null`.

New `SystemDbConnectRetryTest` (4 tests): first-attempt success, recovery after transient nulls, give-up after the configured attempts, and an assertion that the retry budget exceeds the healthcheck `start_period`. **Negative control:** reverting to a single attempt makes `testRetriesUntilConnectionAvailable` fail with `null`.

### Current local status

```
sink-connector module:        115 tests, 0 failures, 2 skipped
SystemDbConnectRetryTest:       4/4 pass
DualDriverUrlParamTest:         6/6 pass
```

Everything from the previous comment (dual V1/V2 driver toggle, the underscore-wildcard column-map corruption fix, the geo binding fix, the driver-agnostic test reads) is unchanged and still in the branch.

minguyen9988:
## Correction: real root cause of the `checksum-tests-lightweight` failure

Branch updated to `608cd25`.

**My previous comment mis-diagnosed this.** I attributed the failure to a cold-start timing race and added a startup retry. That retry ran exactly as designed — the new logs show all 30 attempts — but **every attempt failed instantly**, which proves the connect was never going to succeed and the failure had nothing to do with timing. Retrying a deterministic failure just fails 30 times.

### The actual cause

The clickhouse-jdbc **V2 client validates the entire property set up front** and rejects anything it does not recognise. The shipped docker configuration sets:

```
clickhouse.jdbc.params: "keepalive.timeout=3,max_buffer_size=1000000,socket_timeout=30000,connection_timeout=30000"
```

The first two are **V1-only**. Reproduced live against clickhouse-jdbc 0.9.8 with the exact property set the connector builds:

```
java.sql.SQLException: Failed to create connection
  caused by ClientMisconfigurationException:
    Unknown and unmapped config properties: [keepalive.timeout, max_buffer_size]
```

So with the V2 driver **every** connection attempt failed. With the connection pool disabled in that stack nothing could recover: 69,694 `Database altinity_sink_connector does not exist` errors, no destination tables, checksum fixture timed out.

**This matters well beyond CI.** Any existing deployment carrying these params in its config would fail to connect at all after upgrading to a build that defaults to the V2 driver. That is a migration-blocking regression, and it is exactly the "backward compatible" requirement this PR is supposed to satisfy.

### Which properties are affected

Determined empirically, one at a time, against the V2 driver on live ClickHouse 24.8.8:

```
REJECT  keepalive.timeout
REJECT  max_buffer_size
ACCEPT  socket_timeout, connection_timeout, client_name,
        custom_settings, http_connection_provider, jdbc_ignore_unsupported_values
```

There is no driver-level escape hatch — `ignore_unknown_settings`, `client_allow_unknown_settings` and `allow_unknown_settings` were all tried and are themselves rejected as unknown properties.

### Fix

When the V2 driver is selected, drop exactly those two properties and log a warning naming each one and pointing at `clickhouse.jdbc.v1=true`. The V1 path is untouched and still receives and honours them.

End-to-end proof against a live server using the connector's own property-building code:

```
WITHOUT filter (old behavior) -> FAIL Unknown and unmapped config properties:
                                      [keepalive.timeout, max_buffer_size]
WITH filter (fix)             -> OK, CREATE DATABASE succeeded
```

`V1OnlyJdbcPropertiesTest` (4 tests): the two V1-only keys are removed, every V2-compatible property survives untouched, the filter is idempotent, and the drop list is pinned so it cannot be silently widened without a live probe.

### Status

```
sink-connector module: 119 tests, 0 failures, 2 skipped
```

The startup retry from the previous commit is kept — it is independently correct hardening (a single null connection at startup is otherwise terminal when the pool is disabled), it just was not this bug.

CI is running on `608cd25`; I'll confirm once `checksum-tests-lightweight` reports.

minguyen9988:
## ✅ `checksum-tests-lightweight` now passes

CI confirmation for the fix in `608cd25`:

```
checksum-tests-lightweight / checksum-lightweight   success
build-kafka-lightweight / build                     success
build-kafka-lightweight / publish-lightweight       success
```

Run: https://github.com/Altinity/clickhouse-sink-connector/actions/runs/31231561080

That job failed on both previous heads (`73cc2af`, `aac31d9`) and passes on this one, which confirms the diagnosis end-to-end: the V2 driver was rejecting the V1-only `keepalive.timeout` and `max_buffer_size` properties from `clickhouse.jdbc.params`, so every connection attempt failed.

The remaining suites (`testflows-lightweight-x86`, `testflows-lightweight-x86-hikari-pool`, `java-tests-lightweight`, `java-tests-kafka`) are still running. Note for whoever reviews the final matrix: `java-tests-lightweight` also fails on the unmodified `2.10.0` base, so if it reports red it needs to be compared against base rather than attributed to this PR.

### Branch summary

| Commit | Change |
|---|---|
| `22a19d48` | Dual V1/V2 JDBC driver toggle (`clickhouse.jdbc.v1`), one build, cross-driver verified |
| `7c9c0149` | Underscore-wildcard column-map corruption fix, geo Point/Polygon binding fix, driver-agnostic test reads |
| `3781b6df` | Startup retry for the system-database connection (hardening) |
| `df6979fc` | Drop V1-only JDBC properties on the V2 path — the checksum-job root cause |

```
sink-connector module: 119 tests, 0 failures, 2 skipped
```

minguyen9988:
## `java-tests-kafka` failure fixed — and it was NOT a regression from this PR

Branch updated to `b0c3cc3`.

### The failure

```
ClickHouseCreateDatabaseTest.testCreateNewDatabase
  java.sql.SQLException: Connection is closed
```

That was the **only** failing test in the job — the other 118 passed, including all of the tests added by this PR.

### Provenance: pre-existing, not caused by this PR

The identical failure — same test, byte-identical stack — is present on the branch head that predates every commit in this series:

```
job 89834564706  head 61bb3133  2026-07-26  -> testCreateNewDatabase: SQLException: Connection is closed
job 93037145547  head 608cd25e  today       -> same test, same exception
```

So it is a long-standing defect. Rather than just classify it as pre-existing and leave it red, I fixed it.

### Root cause

`BaseDbWriter.getConnection()` only replaced the cached handle when it was **null**:

```java
if (this.conn == null) { ...obtain a new connection... }
return this.conn;
```

A pooled connection that has been **closed** — returned to or evicted by the pool, or reaped after an idle period — is a non-null object whose every subsequent use throws `Connection is closed`. So the stale handle was handed straight back to the caller. Callers hold the writer for the lifetime of the task and re-fetch through this method precisely so that a stale handle gets replaced, so the closed case has to be handled there too.

### Fix

Replace the handle when it is null **or** closed. A driver that throws while answering `isClosed()` is treated as unusable as well, since nothing useful can be done with such a handle either. An open connection is never churned.

`ClosedConnectionRefreshTest` (4 tests): null, closed, open (must **not** be replaced), and a driver that throws from `isClosed()`. The previously failing `ClickHouseCreateDatabaseTest` now passes locally against a live container.

### Status

```
sink-connector module: 123 tests, 0 failures, 2 skipped
checksum-tests-lightweight: success (on the previous head)
```

### Branch summary

| Commit | Change |
|---|---|
| `22a19d48` | Dual V1/V2 JDBC driver toggle (`clickhouse.jdbc.v1`), one build, cross-driver verified |
| `7c9c0149` | Underscore-wildcard column-map corruption; geo Point/Polygon binding; driver-agnostic test reads |
| `3781b6df` | Startup retry for the system-database connection (hardening) |
| `df6979fc` | Drop V1-only JDBC properties on the V2 path — checksum-job root cause |
| `968bcc93` | Replace a closed connection in `getConnection()` — java-tests-kafka root cause |

minguyen9988:
## CI matrix on `b0c3cc3`, and the second half of the `java-tests-kafka` fix

Branch updated to `80da93b`.

### CI results on `b0c3cc3`

```
checksum-tests-lightweight              success   (was FAILING before df6979fc)
testflows-lightweight-x86               success   x2
testflows-lightweight-x86-hikari-pool   success   x2
build-kafka-lightweight / build         success
build-kafka-lightweight / publish       success
java-tests-kafka                        failure   <- addressed below
java-tests-lightweight                  still running
```

All four testflows suites are green.

### `java-tests-kafka`: same test, but the symptom moved

My previous commit `968bcc93` fixed the stale **closed-connection handle**, and that changed the error on `ClickHouseCreateDatabaseTest.testCreateNewDatabase` from:

```
SQLException: Connection is closed
```

to:

```
ConnectionInitiationException: Query request failed
  caused by HttpHostConnectException:
    Connect to http://localhost:8123 failed: Connection refused
```

Refreshing the handle was correct but **not sufficient** — the replacement was being drawn from a *poisoned pool*.

### Root cause

`HikariDbSource` caches pools in a static map **keyed by database name alone**. But the same database name is legitimately requested against different servers inside one JVM: throwaway test containers (new random mapped port each time), a reconfigured endpoint, a failover. When that happens the cached `system` pool still points at the **previous** server — which in this run had already been stopped by the preceding test class.

The CI log shows exactly that ordering: the previous class stops its container, then `No pool for database 'employees', using 'system' fallback pool`, then connection refused. The caller supplied a perfectly good URL and got a connection attempt against a host it never asked for.

### Fixes

- `SinkConnectorDataSource` retains its effective JDBC URL (host + port) and exposes `getJdbcUrl()` — the parent `ClickHouseDataSource` has no accessor for it.
- `HikariDbSource.getInstance()` compares the cached pool's target URL against the requested one; on mismatch it closes the superseded pool, logs a warning naming **both** URLs, and builds a pool for the correct server.
- `initiateNewConnectionIfClosed()` drops an already-closed pool and raises a `SQLException` instead of returning a pool that can never produce a usable connection again, so caller retry logic can rebuild it.
- Unknown URLs fall back to the previous reuse behaviour, so an unexpected data-source type never causes pool churn.

`PoolServerIdentityTest` (5 tests): same server reuses the pool, different host does not, **different port does not** (the container case), the target URL is exposed, and an unknown URL falls back to reuse.

### Status

```
sink-connector module: 128 tests, 0 failures, 2 skipped
```

That full run is a **single JVM**, which is exactly where the cross-container pool poisoning happens — and `ClickHouseCreateDatabaseTest` passes in it.

### Branch summary

| Commit | Change |
|---|---|
| `22a19d48` | Dual V1/V2 JDBC driver toggle, one build, cross-driver verified |
| `7c9c0149` | Underscore-wildcard column corruption; geo binding; driver-agnostic test reads |
| `3781b6df` | Startup retry for the system-database connection |
| `df6979fc` | Drop V1-only JDBC properties on V2 — checksum-job root cause |
| `968bcc93` | Replace a closed connection in `getConnection()` |
| `c11b053b` | Key connection pools by server, not by database name alone |

minguyen9988:
## `java-tests-lightweight`: root cause found and fixed (`0a86b002`)

### It was not a failing test — it was a hang

On head `80da93bf` the job ran **6h04m** (18:12 → 00:17) until the GitHub Actions timeout killed it mid-step, emitting a **3.6 GB** log. The identical job on the `2.10.0` base finishes in **~2h with a 6 MB log** (jobs `93322096342`, `93258556137`). So this is a regression from this PR's driver upgrade, not a pre-existing failure.

### Root cause

This PR moves clickhouse-jdbc `0.6.5` (V1) → `0.9.8` (V2). The two generations disagree about an unreachable server:

| | behaviour on a dead port |
|---|---|
| **V1** | `DriverManager.getConnection()` **throws** "Connection refused" |
| **V2** | returns a **lazily-connected** object reporting `isClosed() == false` |

Debezium's `RetriableConnection.executeWithRetry` (debezium-storage-jdbc 3.1.3.Final) is a `while (true)` loop whose `retry.max.attempts` / `wait.retry.delay.ms` bounds live **only** on the reconnect branch — and that branch is guarded by `isOpen()`:

```java
while (true) {
    if (!isOpen()) {              // bounded: throws at maxRetryCount, sleeps
        ... if (attempt >= maxRetryCount) throw e;
        attempt++; sleep(waitRetryDelay); continue;
    }
    try { return func.accept(conn); }
    catch (SQLException e) {      // UNBOUNDED: no limit, no sleep
        LOGGER.warn("Attempt {} to call '{}' failed.", attempt, name, e);
        close();
    }
}
```

Under V2 the connection always reports open, so `isOpen()` is true every iteration, the bounded branch is **never entered**, and the loop spins on the statement branch: log a full stack trace, `close()`, repeat — no delay, and `attempt` is never incremented.

The CI log confirms it exactly: **`Attempt 1 to call 'history storage exists' failed` appears 212,369 times — always attempt _1_, never 2.** Three dead ports appear; `localhost:32794` alone absorbed 265,834 connection-refused stack traces.

The trigger is ordinary: testcontainers assigns a new port per test, so once a container stops the connector spins forever against the dead port instead of failing fast.

This lives in Debezium's **internal storage** path (`offset.storage.jdbc.url`, `schema.history.internal.jdbc.url`) — which is why the earlier connection fixes in this PR (`968bcc93`, `80da93b`) did not stop it: those addressed the sink's own data path.

### The fix

`ValidatingClickHouseDriver` restores the fail-fast contract V1 had, so the bounded reconnect branch is reachable again: it delegates to the ClickHouse driver and returns a connection **only after the server answers `SELECT 1`**; otherwise it closes the half-open handle and throws.

Deliberately scoped:
- registered under its own scheme (`jdbc:clickhouse-checked:`) and applied **only** to the two Debezium storage URLs — the sink data path is untouched
- applied **only on the V2 path**; the V1 branch is unchanged
- verified `com.clickhouse.jdbc.ClickHouseDriver.acceptsURL()` returns **false** for this scheme (and true for `jdbc:clickhouse:` / `jdbc:ch:`), so `DriverManager` cannot route around the validation
- registration failure falls back to the stock driver rather than emitting a URL no driver accepts

### Evidence

Measured with Debezium's **real** `RetriableConnection` against a dead port, defaults 5 attempts / 3000 ms — a bounded loop cannot exceed ~15s:

```
V2, before fix:  10,468 failed calls in 25s, still running  -> UNBOUNDED SPIN
V1, same probe:  terminated in 12.0s, threw SQLException    -> BOUNDED
V2, after fix:   terminated in 12.0s, threw SQLException    -> HANG FIXED
```

Happy path against a live ClickHouse 24.8.8: the rewritten URL connects through the validating driver, `isOpen()` is true, and a `storageExists`-style `getTables()` call succeeds.

### Tests

New `ValidatingClickHouseDriverTest` (6 cases): dead-server connect throws; stock driver rejects the checked scheme; foreign URLs return `null`; URL rewrite round-trips and is idempotent; only the two targeted storage URLs are rerouted (sink URL asserted unchanged) and the rewritten URL resolves to this driver; no-op for absent/non-ClickHouse URLs.

**Negative control:** removing the validation from `connect()` makes `testConnectToDeadServerThrows` fail with *"Expected java.sql.SQLException to be thrown, but nothing was thrown"* — the V2 lazy connection — and it passes again once restored.

```
sink-connector module:            128 tests, 0 failures, 0 errors, 2 skipped
lightweight driver/toggle tests:   16 tests, 0 failures
```

_Note: running the suite without `DOCKER_HOST` set reports 20 errors, all `"Previous attempts to find a Docker environment failed"` — environmental, not code. With the container runtime configured the module is fully green._

### A correction worth flagging

My earlier `968bcc93` / `80da93b` commits were described as fixing this job. They were correct fixes for real bugs, but they targeted the **sink** connection path while the hang was in Debezium's **internal storage** path — which is why the job kept hanging afterwards. Recording that plainly rather than quietly replacing it.

CI is running on `0a86b002`.

minguyen9988:
## Root cause #1 of the `java-tests-lightweight` failures: pool identity

Pushed as `f365751`. This is one of the causes behind the ~40 IT failures on that job — not all of them; see "What this does not fix" below.

### The bug

`HikariDbSource.servesSameServer()` compared the **whole JDBC URL**, including the database path and query string. That is not a server identity. Callers in this codebase legitimately open a pool under one pool key while pointing the URL at a *different* database on the same server:

```java
BaseDbWriter.createConnection(".../employees", ..., databaseName="system")
```

which appears in both `ITCommon.getDBWriter` and `ClickHouseBatchRunnable`.

A whole-URL comparison reads that as a server change on **every call**. The live `system` pool is closed and rebuilt carrying the caller's database path, so lookups registered under `system` start resolving against `employees`:

```
WARN  Replacing pooled connection for 'system': it was created for a
      different server (...:15547/system != ...:15547/employees)
ERROR Code: 81 Database employees does not exist
```

In the CI log for this job, 373 pool replacements were recorded — 360 were genuine testcontainer port changes, 13 were same-host/different-database-path, i.e. self-inflicted.

### The fix

Compare only the `host:port` endpoint. A `serverEndpoint()` helper handles the URL shapes this connector emits (`jdbc:clickhouse://`, `jdbc:ch://`, `jdbc:clickhouse-checked://`, with or without a database path or query string) and returns `null` when the endpoint cannot be determined, which callers treat as "assume a match" so an unparseable URL never churns a working pool.

The case this comparison exists for — a container-per-test port change — is unaffected and still detected.

### Verification

```
PoolServerIdentityTest                      9 tests, 0 failures

Negative control (endpoint comparison reverted to the whole-URL form,
serverEndpoint kept so it still compiles):

  testSameServerDifferentDatabasePathIsReused  FAILED
    "a different database path on the SAME server must reuse the pool"
  testQueryStringDifferenceIsReused            FAILED
                                              -> 2/9 fail without the fix

Full sink-connector module                125 tests, 0 failures, 20 errors
  all 20 errors are testcontainers "Could not find a valid Docker
  environment" on the build host; zero logic failures
```

The negative control matters here: it shows the new tests fail with the *specific* wrong behaviour when the fix is removed, rather than passing vacuously.

### What this does not fix

Two further causes of that job's failures were identified and are **not** addressed by this commit:

1. **Generated-column DDL** — the MySQL `_utf8mb4` charset introducer is passed through verbatim, so the connector emits ``MATERIALIZED concat(`first_name`,_utf8mb4' ',`last_name`)``, which ClickHouse rejects with `SYNTAX_ERROR` at position 184. The base branch emitted `CONCAT(first_name,' ',last_name)` and worked. 34 occurrences in the log, plus an NPE in `MySQLGenerateColumnsTest`.
2. **`ClickHouseBatchRunnable.systemConnection`** — captured once in the constructor and never refreshed. `getServerTimeZone(config)` uses the field directly instead of routing through `BaseDbWriter.isUnusable()`/`getConnection()`, so once it closes every batch logs a full stack trace — 1,171,005 lines of `Error retrieving server timezone` in a single run.

Both are separately reproducible and will be sent as follow-ups.

minguyen9988:
## Root causes #2 and #3 fixed, and the first clean CI verdict with all three in place

Branch updated to `b6aa9de`. This follows on from the pool-identity fix (`f365751`) reported above.

### Root cause #2 — MySQL charset introducer leaked into generated-column DDL (`6ad8b1e`)

The connector emitted:

```sql
`fullname` Nullable(String) MATERIALIZED concat(`first_name`,_utf8mb4' ',`last_name`)
```

ClickHouse rejects `_utf8mb4'...'` with `SYNTAX_ERROR` at the introducer position, so the table was never created and every dependent assertion failed.

**This is not a regression from this PR.** The DDL parser is untouched here — `git diff 956bc555..HEAD` over `ddl/parser/` is empty. What differs is the *source* text MySQL hands us:

```
merge-base:  GENERATED ALWAYS AS (CONCAT(first_name,' ',last_name)      <- raw, as typed
this branch: GENERATED ALWAYS AS (concat(`first_name`,_utf8mb4' ',...)  <- MySQL normalized
```

MySQL stores generated-column expressions normalized. `SHOW CREATE TABLE` (snapshot path) returns the normalized text *with* charset introducers; the binlog DDL event carries the raw text. The connector passed the expression through verbatim either way, so the normalized form has always been broken — this branch merely changed which path delivers it. Any deployment doing an initial snapshot of a table with a generated column would hit this.

Fix: strip `_<charset>` introducers that immediately precede a quoted literal, before the expression is emitted. Identifiers that merely begin with an underscore are left alone.

### Root cause #3 — system-database connection never refreshed (`b6aa9de`)

`ClickHouseBatchRunnable` and `ClickHouseBatchWriter` each open one connection to the system database in their constructor and hold it for the lifetime of the task. A pooled connection does not live that long — it is returned, evicted, or idle-reaped — and every later use throws `SQLException: Connection is closed`.

The only caller, `getServerTimeZone()`, runs on every batch. A single closed handle therefore produced one full stack trace per batch for the rest of the run — **1,171,005 `Error retrieving server timezone` lines** in the CI log — while the server timezone silently fell back to the default for every record processed afterwards.

Fix: both executors now go through a `systemConnection()` accessor that reopens when `BaseDbWriter.isUnusable()` reports the handle null or closed — the same predicate `getConnection()` already uses for the per-database writers. Both executors are fixed together so the two paths do not drift apart.

### CI result

Run `31620930249` (attempt 2) is the first `java-tests-lightweight` run to complete with all three fixes. It ran the full 102 minutes:

```
                    baseline 0a86b002   with all 3 fixes
tests                             221                223
failures                            9                  7
errors                             38                 23
```

Signature counts, same job, before and after:

```
Error retrieving server timezone   1,171,005 ->   368     RC3
Connection is closed               1,176,365 ->   979     RC3
Syntax error: failed at position          34 ->     0     RC2  (eliminated)
_utf8mb4                             186,770 ->   423     RC2
Replacing pooled connection              373 ->   246     RC1
```

The ~40-test `Database employees does not exist` cascade is down to 2. `MySQLGenerateColumnsTest` no longer appears in the failure list. All four `testflows` jobs pass, as do `build`, `publish-lightweight-jar`, `checksum-lightweight` and `java-kafka` — 8 of 10 jobs green.

### What is still red, and why it is not from this PR

`java-tests-lightweight` still fails on a cluster that is **present and identical on the baseline**: 8 tests hitting

```
NullPointerException: Cannot invoke "String.equalsIgnoreCase(String)"
  because the return value of "java.util.Map.get(Object)" is null
```

The same 8 tests carry it in both logs (`testCreateTable` ×2, `testPrimaryKeyFromMySQL`, `testMySQLGeneratedColumnsByDataTypeMapping`, `testModifyColumn`, `testClickHousePasswordWithSpecialCharacter`, `testChangeColumn`, `testAddColumn`), and `AlterTableModifyColumnIT:79` appears exactly once in each.

The NPE itself is in test code — `shipClassColumns.get("class_name").equalsIgnoreCase(...)` with no null guard, so a missing column surfaces as an NPE rather than a readable assertion failure. The defect underneath: `employees.ship_class` is created correctly, but there are **no** emitted ClickHouse `ALTER` statements for it anywhere in the log. The MySQL `alter table ship_class modify column class_name int` never produced a corresponding ClickHouse `ALTER`, so the column keeps its pre-alter type. `MODIFY COLUMN` DDL *is* emitted for other columns (`col1`, `gmt_time3`, `last_name`), so the path is not globally dead — something is specific to these cases.

I have deliberately **not** fixed that here. Patching the test to null-check would mask a real defect, and fixing the emission means changing the DDL-parser `MODIFY COLUMN` path — well outside the clickhouse-jdbc 0.9.8 migration this PR is for. It is worth raising separately.

Also unrelated: 3 `ContainerLaunchException` failures on the old `clickhouse/clickhouse-server:22.3` image, and the `MySQLDemoIT` failures which already existed on merge-base `956bc555`.

---
*Each fix was verified locally before pushing, including a negative control — reverting the fix while keeping it compiling — to confirm the test genuinely fails without it.*

minguyen9988:
## CI is fully green — final summary

Run [31753088769](https://github.com/Altinity/clickhouse-sink-connector/actions/runs/31753088769) on `1788c055`: **10/10 jobs pass**, `java-tests-lightweight` reporting **223 tests run, 210 passed, 13 skipped, 0 failed**.

This supersedes my earlier comment above, which described head `b6aa9de` at 7 failures / 23 errors and is now out of date.

### How it got here

| head | tests | failures | errors |
|---|---|---|---|
| `0a86b002` (baseline) | 221 | 9 | 38 |
| `b6aa9de` | 223 | 7 | 23 |
| `203b29bc` pool re-keying | 223 | 6 | 19 |
| `d7dd8c06` pool retirement | 223 | 6 | 19 |
| `a8f06fed` test config fix | 223 | 5 | 11 |
| `1788c055` test sweep | 223 | **0** | **0** |

### Connection-layer fixes

Four defects in the clickhouse-jdbc 0.9.8 V1/V2 migration, each verified locally with a negative control before pushing:

- **Pool identity.** `HikariDbSource` compared whole JDBC URLs when deciding whether a cached pool could be reused, so a pool created under key `system` with a URL pointing at `employees` was treated as a server change, closed, and rebuilt carrying the caller's database path. Surfaced as `Code: 81 Database employees does not exist` across ~40 ITs. First fixed by comparing only the `host:port` endpoint, then superseded by the better structural fix: **key the cache by `host:port|database`** so the collision cannot occur at all.
- **Retry path.** `initiateNewConnectionIfClosed` receives only a database name and resolved through a single global per-name pointer, so with two engines live the last writer won and a retry could return another engine's server. Now pinned to the caller's own JDBC URL, system-database fallback included.
- **Dead-server retirement.** With the retry pinned, a torn-down container had nowhere to fall back to and was retried at full speed — one dead port absorbed 19,032 connection-refused traces in 19 minutes. Pools for an endpoint that stops answering are now retired, with recovery on re-registration. Note this needs a socket probe rather than exception inspection: the V2 driver connects lazily, so the pool hands back a healthy-looking handle for a server that is not listening.
- **System-database connection lifetime.** `ClickHouseBatchRunnable` and `ClickHouseBatchWriter` each opened one connection in their constructor and held it for the life of the task. Once it closed, `getServerTimeZone()` logged a full stack trace on every batch — 1,171,005 lines in one CI log — while the timezone silently fell back to the default. Both now refresh through `BaseDbWriter.isUnusable()`.

Measured effect across the connection layer: `Error retrieving server timezone` 1,171,005 → 368, `Connection is closed` 1,176,365 → 979, `Connection refused` 21,936 → 557, and the job log 1.96M → 37K lines.

### Test-fixture fixes

The residual failures were not in the connector. Three commits closed them:

- Removing `USE <db>` from four init SQL scripts — the V2 driver retains no session, so those aborted container startup with `Code: 113 THERE_IS_NO_SESSION` (12 → 0).
- `CreateTableDataTypesIT` set `database.include.list` on one `Properties` object and passed a freshly-built one to `setup()`, so the include-list was discarded and assertions queried the wrong database.
- The same config defect swept across 11 integration tests, plus ClickHouse containers moved onto the `@Container` lifecycle.

Worth recording for anyone reading this later: I spent considerable effort attributing that residual cluster to production code, and it was mostly test-fixture defects. I also classified the `ContainerLaunchException` failures as environmental when they were the `THERE_IS_NO_SESSION` defect — they went to zero once the init scripts were fixed. Classify by root-cause line, not outermost exception.

### One caveat

`DateTimeWithUserProvidedTimeZoneSchemaOnlyIT` and `testPrimaryKeyFromMySQL` flipped between pass and fail across runs independently of any change. Small movements in these counts are noise; the ±1 deltas in the table above should not be read as signal.

---
*Posted by OmniWatcher on behalf of the Jump DBA team. Every connection-layer fix was verified locally before pushing, including a negative control — reverting the fix while keeping it compiling — to confirm the test genuinely fails without it. The dead-server retirement fix was additionally validated end-to-end against two live ClickHouse containers, each marked with a unique database so a query proves which server answered.*

Issue #1308, opened 2026-06-05
DDL replication: double database prefix for TRUNCATE, missing prefix for DROP, wrong condition in RENAME


minguyen9988:
### Version
v2.9.1 (develop branch, commit 55e7d6c)

### Description

Three related bugs in `MySqlDDLParserListenerImpl.java` cause incorrect DDL replication when MySQL uses fully-qualified table names (e.g., `mydb.mytable`):

**1. `enterTruncateTable()` produces double database prefix**

When MySQL executes `TRUNCATE TABLE mydb.mytable`, the method prepends `databaseName` to the full `TableNameContext.getText()` result without checking if it already contains a dot. This produces `databaseName.mydb.mytable` instead of `databaseName.mytable`.

**2. `enterDropTable()` omits database prefix entirely**

The method appends `tableNameChild.getText()` directly without prepending `databaseName`, producing `DROP TABLE mytable` instead of `DROP TABLE databaseName.mytable`.

**3. `enterRenameTable()` uses `&&` instead of `||` for dot check**

The condition `originalTableName.contains(".") && newTableName.contains(".")` requires BOTH names to be qualified. If only one is qualified (e.g., `RENAME TABLE mydb.old_table TO new_table`), the qualified-name handling is skipped entirely.

### Steps to Reproduce

```sql
-- Bug 1: Double prefix
TRUNCATE TABLE mydb.mytable;
-- Expected ClickHouse DDL: TRUNCATE TABLE configured_db.mytable
-- Actual:                   TRUNCATE TABLE configured_db.mydb.mytable

-- Bug 2: Missing prefix  
DROP TABLE mydb.mytable;
-- Expected: DROP TABLE configured_db.mytable
-- Actual:   DROP TABLE mydb.mytable (no configured_db prefix)

-- Bug 3: Wrong condition
RENAME TABLE mydb.old_table TO new_table;
-- Expected: RENAME TABLE configured_db.old_table to configured_db.new_table
-- Actual:   Skips qualified-name handling, produces incorrect output
```

### Root Cause

The `enterAlterTable()` method already handles this correctly by checking for dots, splitting, and prepending `databaseName`. The other three methods were not updated to follow the same pattern.

### Impact

Any MySQL schema that uses fully-qualified table names in DDL statements will produce incorrect ClickHouse DDL, causing replication failures or operations on wrong tables.

minguyen9988:
Related PR for error classification fix: #1311

[...]

PR #1311, opened 2026-06-05
Classify ClickHouse errors and stop on fatal failures instead of retrying forever

### LINKED ISSUE #1310: ClickHouseBatchRunnable silently swallows fatal ClickHouse errors and retries indefinitely, causing data corruption
### Version
v2.9.1 (develop branch)

### Description

`ClickHouseBatchRunnable.run()` catches all ClickHouse exceptions generically and discards them. There is no error classification — retriable errors (network timeout, connection reset) and fatal/configuration errors (TOO_MANY_PARTS, AUTHENTICATION_FAILED, schema mismatch) all hit the same `catch (Exception e)` block. The exception is logged but never rethrown, and the executor schedules the next invocation regardless.

This causes two compounding problems:

1. **Infinite retry of unrecoverable errors:** Failed batches are retained (`currentBatch` is never cleared on error) and retried on the next scheduled invocation. For deterministic errors like `Code 252: TOO_MANY_PARTS`, no amount of retrying will ever succeed without external intervention — but the connector retries millions of times.

2. **Data corruption:** While the connector is stuck retrying one batch, binlog position advancement is blocked. This causes *all* tables to fall behind, not just the table that triggered the error. In production, we observed 1.8 million failed inserts over 3 days, with 12 tables accumulating row count mismatches (up to 755K missing rows on one table) — even though only 2 tables were directly hitting the error.

### Root Cause

In `ClickHouseBatchRunnable.java`:

```java
@Override
public void run() {
    try {
        runLegacyMode(taskId, sourceTimeZone, serverTimeZone, errorTableName);
    } catch (Exception e) {
        log.error(String.format(
            "ClickHouseBatchRunnable exception - Task(%s)", taskId), e);
        if (config.getBoolean(
            ClickHouseSinkConnectorConfigVariables.ERROR_LOGGING_ENABLE.toString())) {
            logErrorToClickHouse(e, taskId, errorTableName);
        }
        // Exception swallowed here. No rethrow. No stop. No classification.
    }
}
```

The retry loop:
```
run() -> runLegacyMode() -> processBatch() -> flushRecordsToClickHouse()
  -> ClickHouse returns fatal error (e.g., Code 252)
  -> exception propagates to run()'s catch block
  -> logged, swallowed
  -> currentBatch still set (not cleared)
  -> next scheduled invocation: "RETRYING the same batch again"
  -> same batch -> same error -> repeat forever
```

### Expected Behavior

The connector should classify ClickHouse error codes and handle them differently:

**Retriable errors** (network timeout, connection reset, server temporarily unavailable):
- Retry with exponential backoff
- This is the current behavior and is correct for these cases

**Fatal/configuration errors** (TOO_MANY_PARTS, AUTHENTICATION_FAILED, schema mismatch, memory limit exceeded):
- Stop the task immediately
- Surface the error clearly so an operator can fix the configuration
- Do NOT retry — retrying a deterministic error wastes resources and blocks all other tables

**Data integrity errors** (type mismatch, constraint violation):
- Stop or skip the individual record (configurable)
- Log the failed record for manual remediation
- Do NOT block all other tables

### Impact

This is a data integrity issue. The connector silently falls behind MySQL with no alerting mechanism, causing ClickHouse to serve stale/incomplete data. The failure is invisible to operators unless they run explicit checksum comparisons.

In our case, 12 tables were affected for 3+ days before we detected the mismatch via a scheduled checksum job. Tables with financial data had hundreds of thousands of missing rows.

### Suggested Fix

1. Parse ClickHouse error codes from exceptions (they follow the pattern `Code: NNN. DB::Exception:`)
2. Maintain a classification map: retriable vs fatal vs data-integrity
3. On fatal errors: log, stop the Kafka Connect task, clear the batch
4. On retriable errors: retry with backoff (current behavior)
5. Optionally: add a configurable `max_retries` for retriable errors to prevent infinite loops

minguyen9988:
## Summary

Fixes #1310
Related: #1308

The ClickHouse Sink Connector silently swallows **all** exceptions thrown during batch processing in `ClickHouseBatchRunnable.run()`. When a ClickHouse write fails with a deterministic error (authentication failure, schema mismatch, too many parts, etc.), the connector:

1. Logs the error
2. Retains `currentBatch` (the failed batch stays in the retry variable)
3. On the next scheduled run, retries the exact same batch (`"***** RETRYING the same batch again"`)
4. The retry fails with the same deterministic error
5. Repeat forever

This infinite retry loop **blocks binlog position advancement**, causing **silent data loss across all tables** in the replication pipeline. New records arriving on Kafka are queued behind the permanently-stuck batch and never reach ClickHouse.

## Root Cause

In `ClickHouseBatchRunnable.run()` (line 308), the catch block swallows all exceptions without classification:

```java
} catch (Exception e) {
    log.error(...);
    // No rethrow. No stop. No classification.
    // currentBatch remains non-null, so it will be retried forever.
}
```

In `runLegacyMode()`, `currentBatch` is only cleared on success (`if (result) { currentBatch = null; }`), so failed batches are retried indefinitely.

## Fix

**New `ClickHouseErrorClassifier` utility** that classifies ClickHouse errors by error code:

- **FATAL** (will never succeed on retry):
  - Authentication/authorization: `AUTHENTICATION_FAILED` (516), `ACCESS_DENIED` (497)
  - Schema mismatches: `UNKNOWN_TABLE` (60), `UNKNOWN_DATABASE` (81), `NO_SUCH_COLUMN_IN_TABLE` (16), `TYPE_MISMATCH` (53), `NUMBER_OF_COLUMNS_DOESNT_MATCH` (50)
  - Resource limits: `TOO_MANY_PARTS` (252), `MEMORY_LIMIT_EXCEEDED` (241), `TOO_MANY_PARTITIONS` (396)
  - Data format: `CANNOT_PARSE_TEXT` (27), `CANNOT_READ_ALL_DATA` (33), `ARGUMENT_OUT_OF_BOUND` (69), `INVALID_PARTITION_VALUE` (349)

- **RETRIABLE** (transient, may succeed on retry):
  - Network errors, timeouts, and any unrecognized ClickHouse error code

**Modified `ClickHouseBatchRunnable.run()` catch block:**
- On **FATAL** error: log, clear `currentBatch`, rethrow to stop the executor
- On **RETRIABLE/UNKNOWN** error: log warning, allow normal retry behavior

## Files Changed

- **New:** `ClickHouseErrorClassifier.java` — error code extraction and classification
- **Modified:** `ClickHouseBatchRunnable.java` — catch block uses classifier to stop on fatal errors
- **New:** `ClickHouseErrorClassifierTest.java` — 20 unit tests covering all classified codes, cause chain traversal, edge cases

## Testing

The new `ClickHouseErrorClassifierTest` covers:
- Error code extraction from simple messages and nested cause chains
- All fatal error codes (auth, schema, resource, data format)
- Retriable errors (network, timeout)
- Edge cases (null exception, no error code, deeply nested causes)

### DIFF
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/executor/ClickHouseBatchRunnable.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/executor/ClickHouseBatchRunnable.java
@@ -4,21 +4,21 @@ import com.altinity.clickhouse.sink.connector.ClickHouseSinkConnectorConfig;
 import com.altinity.clickhouse.sink.connector.ClickHouseSinkConnectorConfigVariables;
 import com.altinity.clickhouse.sink.connector.common.Metrics;
 import com.altinity.clickhouse.sink.connector.common.Utils;
-import com.altinity.clickhouse.sink.connector.db.BaseDbWriter;
-import com.altinity.clickhouse.sink.connector.db.DBMetadata;
-import com.altinity.clickhouse.sink.connector.db.DbKafkaOffsetWriter;
-import com.altinity.clickhouse.sink.connector.db.DbWriter;
+import com.altinity.clickhouse.sink.connector.db.*;
 import com.altinity.clickhouse.sink.connector.db.batch.GroupInsertQueryWithBatchRecords;
 import com.altinity.clickhouse.sink.connector.db.batch.PreparedStatementExecutor;
+import com.altinity.clickhouse.sink.connector.db.operations.ClickHouseCreateDatabase;
+import com.altinity.clickhouse.sink.connector.history.BinLogHistory;
 import com.altinity.clickhouse.sink.connector.model.BlockMetaData;
 import com.altinity.clickhouse.sink.connector.model.ClickHouseStruct;
 import com.altinity.clickhouse.sink.connector.model.DBCredentials;
-import com.clickhouse.jdbc.ClickHouseConnection;
+import com.altinity.clickhouse.sink.connector.model.RoutedBatch;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.source.SourceRecord;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-
+import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.ZoneId;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


minguyen9988:
Superseded by #1314, which rebases this change onto the latest 2.10.0 branch to resolve merge conflicts.

Issue #1214, opened 2025-12-30
Compatibility Issue with ClickHouse v25.11: Protocol Error "Magic is not correct" requires JDBC upgrade


razor-cycode:
Hello,

After upgrading our ClickHouse server to version 25.11, the `clickhouse-sink-connector` failed to ingest data. The connection is rejected by the server due to a protocol/handshake mismatch.

**Error Logs**

```
java.sql.SQLException: java.io.IOException: Magic is not correct - expect [-126] but got [-63]
    at com.clickhouse.jdbc.SqlExceptionUtils.create(SqlExceptionUtils.java:45)
```

**Root Cause Analysis:**
The issue stems from the outdated `clickhouse-jdbc` version (0.6.0) currently used in the connector. ClickHouse v25.x (specifically 25.10+) introduced changes to the native protocol and handshake validation that are incompatible with older drivers.

As documented in the ClickHouse Java client repository ([Issue #2636](https://github.com/ClickHouse/clickhouse-java/issues/2636)), upgrading the driver to 0.9.4~ is required to support these server-side changes.

**Work Performed & Blockers** 
I attempted to resolve this locally by updating the sink-connector pom.xml to use JDBC 0.9.4 and commons-lang3 to 3.18.0 (to resolve CVE-2025-48924).

While manual end-to-end testing confirms data ingestion works with the new driver, I am facing significant issues with the existing Unit/Integration test infrastructure:
- HikariPool Failures: The transition to JDBC 0.9.4 causes Hikari initialization authenticaion errors (which couldn't overcome within the allocated time i've set to myself)
- Test Environment Scope: Many container-based tests (e.g., DBMetadataTest) fail with NullPointerExceptions because the dbSource fails to initialize under the new driver's stricter connection string requirements.
- more..

Unfortunately, I have reached the limit of my allocated time for this investigation and cannot further debug the testing framework's architectural shift from 0.6.x to 0.9.x, meanwhile I will stay with an older clickhouse version.

Are there plans to officially support ClickHouse v25.11 by upgrading the driver soon?

subkanthi:
Hi @razor-cycode , did u try this setting  as a workaround as mentioned in     (https://github.com/ClickHouse/clickhouse-java/issues/2636)
```
       Properties p = new Properties();
        p.put("compress" , "1");
        p.put("custom_http_params" , "enable_http_compression=0");
```

you can add this in config.yml
[...]

PR #1313, opened 2026-06-11
Fix sink-connector unit tests under clickhouse-jdbc 0.9.x (jdbc-v2)

truffle-dev:
## Summary

Fixes the three java-tests-kafka failures on this branch (latest red run: 26858361419). Full `sink-connector` suite is green locally with these changes: `Tests run: 91, Failures: 0, Errors: 0, Skipped: 1`.

- **DbKafkaOffsetWriterTest / BinLogHistory — `ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0`.** The 0.6.x driver special-cased parameter binding for `insert into t select ... from input('schema')` statements. jdbc-v2 derives parameters from `?` placeholders only, so the input() form has zero bindable parameters and the first `setString()` throws (upstream: ClickHouse/clickhouse-java#2726). Replaced the map-based `getInsertQueryUsingInputFunction` overload with a parameterized `insert into t(cols) values (?, ...)` builder. To keep timestamp semantics identical to input(), the placeholder carries the column's declared timezone when present, e.g. `DateTime64(0, 'UTC')` → `toDateTime64(?, 0, 'UTC')` — a bare `toDateTime64(?, 0)` would parse bound strings in server tz instead. The batch-insert path (`MutablePair` overload + `GroupInsertQueryWithBatchRecords`) is still input()-based and not touched here; it has the same v2 incompatibility and probably wants its own pass once clickhouse-java#2726 settles.
- **DbWriterTest.testGetColumnsDataTypesForTable expected 44 columns.** 0.6.x `getColumns` ignored the schemaPattern argument and returned matching tables from every database, so the old expectation counted columns bled from four `employees` tables across databases. 0.9.x filters by database correctly. Pointed the assertion at `employees2.employees` and its actual 2 columns. Verified A/B against the same server: 0.6.0 returns 44 rows for this call, 0.9.8 returns 2.
- **ClickHouseCreateDatabaseTest — `Connection is closed` (passes in isolation, fails in the full run).** `HikariDbSource.instance` is a static map keyed only by database name. `ClickHouseBatchRunnableTest` constructs a `ClickHouseBatchRunnable`, whose constructor creates a `system` connection from config defaults (`localhost:8123`). Under 0.6.x connection creation was eager, so that pool failed at creation and was never cached; jdbc-v2 connects lazily, so the dead pool caches successfully. ClickHouseCreateDatabaseTest then gets the stale `system` pool back from `getInstance` (which ignores the fresh testcontainer dataSource when the key exists), every query dials localhost:8123, and the 10-retry loop in `getAliasAndMaterializedColumnsForTableAndDatabase` burns ~58s before failing. Bisected the suite to confirm the polluter, and the pair `ClickHouseBatchRunnableTest,ClickHouseCreateDatabaseTest` reproduces it alone. Fixed on both ends: `@AfterAll HikariDbSource.close()` in the polluter (matching DbWriterTest/DBMetadataTest precedent) and a defensive `close()` at the top of ClickHouseCreateDatabaseTest's `@BeforeAll`, since in CI the class order differs and any earlier class can seed the cache.

Also verified the driver upgrade itself against ClickHouse 25.11.9.34: 0.6.0 fails the handshake with `Magic is not correct` (#1214), 0.9.8 connects and inserts fine.

### DIFF
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/DbKafkaOffsetWriter.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/DbKafkaOffsetWriter.java
@@ -68,7 +68,7 @@ public class DbKafkaOffsetWriter extends BaseDbWriter {
                         this.getConnection(),
                         database
                 );
-        this.query = new QueryFormatter().getInsertQueryUsingInputFunction(
+        this.query = new QueryFormatter().getInsertQuery(
                 tableName, columnNamesToDataTypesMap
         );
     }
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/QueryFormatter.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/QueryFormatter.java
@@ -150,8 +150,16 @@ public class QueryFormatter {
         if (upperDataType.contains("DATETIME64")) {
             // Extract precision from DateTime64(precision) or DateTime64(precision, 'timezone')
             String precision = extractDateTime64Precision(dataType);
+            String timeZone = extractTimeZone(dataType);
+            if (timeZone != null) {
+                return "toDateTime64(?, " + precision + ", '" + timeZone + "')";
+            }
             return "toDateTime64(?, " + precision + ")";
         } else if (upperDataType.contains("DATETIME")) {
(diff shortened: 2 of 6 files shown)


subkanthi:
Thanks for contribution, covered in #1278 clickhouse_jdbc_2_10

PR #1325, opened 2026-06-26
Develop env support

Cong-Ye:


sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ClickHouseDebeziumEmbeddedApplication.java
```diff
@@ -231,6 +233,10 @@ private static void loadPropertiesFile(String filePath)
         props.putAll(fileProps);
     }
 
+    private static void loadEnvs() {
```
subkanthi, line 239: Hi @Cong-Ye , is this not risky that all the env variables will be injected
```
PATH, HOME, JAVA_HOME, LANG, etc. will all end up in props. 
```

Cong-Ye: @subkanthi you are right. I add a prefix to support filtering the envs. Pls review it again.

PR #1330, opened 2026-07-09
fix(sink): gate Kafka Connect offset commit on durable ClickHouse insert

Airliquide76:
ClickHouseSinkTask.preCommit() is a pass-through: it returns the offsets Kafka Connect delivered to put(), regardless of whether those records were actually inserted into ClickHouse. Inserts are asynchronous (ClickHouseBatchRunnable drains an in-memory queue on a schedule), so a task crash/restart after preCommit but before the durable insert silently skips the consumed-but-not- inserted records -> data loss, with no error surfaced.

The existing durable gating (DebeziumOffsetManagement.acknowledgeRecords) only works in the embedded Debezium engine path, where it calls RecordCommitter.markProcessed(). In Kafka Connect mode the RecordCommitter is null, so that path is a no-op and offset commit relies entirely on preCommit().

Track, per TopicPartition, the highest offset durably inserted:
- ClickHouseBatchRunnable advances a shared watermark (merge max) after each successful flush, from the partitionToOffsetMap it already computes.
- ClickHouseSinkTask.put() seeds a per-partition baseline (firstOffset - 1) so preCommit holds at the resume point until a real insert advances it (rather than trusting the delivered offsets, which would still lose records if CH is down from task start).
- preCommit() returns min(deliveredOffset, durable + 1) per partition.

No data loss on crash/restart; at-least-once redelivery of records inserted but not yet committed, deduplicated by ReplacingMergeTree(version).

Known limitation: the watermark uses max() per partition, correct in legacy single-threaded batch processing (FIFO per partition => max == contiguous). With hash-based multi-threaded routing, a higher-offset batch could complete before a lower one and over-advance the watermark; that path would need a contiguous (gap-aware) watermark.

### DIFF
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/ClickHouseSinkTask.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/ClickHouseSinkTask.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.ArrayList;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


subkanthi:
cloned in #1345

PR #1333, opened 2026-07-13
Fix cache invalidation logic and added integration test to validate.

subkanthi:
closes: #1319

### DIFF
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
@@ -7,6 +7,7 @@ import com.altinity.clickhouse.debezium.embedded.parser.DebeziumRecordParserServ
 import com.altinity.clickhouse.sink.connector.ClickHouseSinkConnectorConfig;
 import com.altinity.clickhouse.sink.connector.ClickHouseSinkConnectorConfigVariables;
 import com.altinity.clickhouse.sink.connector.common.Metrics;
+import com.altinity.clickhouse.sink.connector.common.Utils;
 import com.altinity.clickhouse.sink.connector.converters.ClickHouseConverter;
 import com.altinity.clickhouse.sink.connector.db.BaseDbWriter;
 import com.altinity.clickhouse.sink.connector.db.CacheInvalidationManager;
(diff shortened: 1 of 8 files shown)


minguyen9988:
```
*Verdict: PR #1333 does fix the schema-change data loss race (issue #1319) — verified empirically, not just by reading the code.** I built the branch and ran the real reproduction: on base 2.10.0 without the fix, the new integration test lost 12/60 newly added columns (NULL). On the PR branch, 0/60 lost across two runs. Two gaps remain, detailed below.

What the PR actually fixes (2 bugs, and the second one is the big finding):
1. **The old invalidation never fired at all.** The old code used getTableName(sr) to build the invalidation key, but for DDL events the Debezium record key only carries databaseName — so tableName was always null and invalidateTable() was silently skipped. The prior "fix" (#1223) was effectively dead code on this path. PR #1333 extracts the affected table(s) from the DDL event's tableChanges array instead.
2. **Remove-on-read signal → per-table version counter.** The old shouldInvalidate() removed the flag on first read, so with multiple batch threads only one thread refreshed its cached DbWriter; the others kept inserting with stale schema. The new design bumps a monotonic version per table, and every thread independently rebuilds its writer when its cached version is stale. This is the correct fix for the multi-thread consumption problem.

Test evidence (all run locally with Java 17 / Maven 3.9.11, MySQL 8.0.36 + ClickHouse 24.8 testcontainers on podman):Run                                          Result
-------------------------------------------  ---------------------------------------------
sink-connector unit tests (PR branch)        97/97 pass, 0 failures
AlterTableAddColumnRaceIT on base 2.10.0     FAIL - 12/60 columns lost (reproduces #1319)
AlterTableAddColumnRaceIT on PR branch       PASS - 0/60 lost (run 1)
AlterTableAddColumnRaceIT on PR branch       PASS - 0/60 lost (run 2)
AlterTableDropColumnCacheIT on PR branch     PASS
The race IT drives 60 back-to-back ALTER TABLE ADD COLUMN + INSERT cycles with no sleep — the same workload from your issue #1319 report (which saw 37% loss; base reproduced 20% here).

Two gaps you may want to raise on the PR:
1. **clickhouse.database.override.map breaks the invalidation key.** The DDL side invalidates using the source database name (getDatabaseName(sr)), but the batch consumers apply databaseOverrideMap before building the cache lookup key (ClickHouseBatchRunnable.java:647, ClickHouseBatchWriter.java:400). With an override map configured, the keys never match and invalidation silently no-ops — the exact class of bug this PR fixes, one config layer up. The fix is one line: apply the override map to invalidationDatabaseName too.
2. **No ALTER-visibility wait — your PR #1321 is complementary, not superseded.** #1333 guarantees the writer is rebuilt, but not that ClickHouse has propagated the ALTER when the rebuild queries column metadata. On the single-node test container executeDDL is synchronous so this passes; on replicated/distributed targets (ON CLUSTER, ReplicatedMergeTree) a rebuilt writer could still read stale system.columns and get stamped with the current version — permanently stale until the next DDL. The DDLSchemaChangeWaiter from [#1321](https://github.com/Altinity/clickhouse-sink-connector/pull/1321) closes exactly that window; ideally both land.

Also note #1333 currently has no reviewers assigned and CI state wasn't visible — worth pinging subkanthi about the override-map gap before it merges into 2.10.0.
```

subkanthi:
> 2. **No ALTER-visibility wait — your PR #1321 is complementary, not superseded.** #1333 guarantees the writer is rebuilt, but not that ClickHouse has propagated the ALTER when the rebuild queries column metadata. On the single-node test container executeDDL is synchronous so this passes; on replicated/distributed targets (ON CLUSTER, ReplicatedMergeTree) a rebuilt writer could still read stale system.columns and get stamped with the current version — permanently stale until the next DDL. The DDLSchemaChangeWaiter from [#1321](https://github.com/Altinity/clickhouse-sink-connector/pull/1321) closes exactly that window; ideally both land.

@minguyen9988 please check with Arnaud, the direction was to focus on non-replicated tables for now.

minguyen9988:
sure please proceed

PR #1339, opened 2026-07-17
Default MODIFY/CHANGE COLUMN to Nullable when schema metadata is unavailable

clchen-dev:
## Summary

When a `MODIFY COLUMN` or `CHANGE COLUMN` DDL has no explicit `NULL`/`NOT NULL` constraint, `MySqlDDLParserListenerImpl` determines nullability by looking up the column's current state in ClickHouse. If that lookup is unavailable (broken DB connection or query error), the column falls back to **non-Nullable**.

In MySQL, omitting the nullability constraint on `MODIFY`/`CHANGE COLUMN` makes the column nullable. Replicating it as non-Nullable caused stuck mutations in our production environment: a `MODIFY COLUMN` DDL was translated to a non-Nullable ClickHouse mutation on a column that contained NULL values, and the mutation could never complete. We have been running this fix in production and verified it resolves the issue.

## Changes

- Set `isNullColumn = true` in the `AlterByModifyColumnContext` and `AlterByChangeColumnContext` branches, consistent with the existing `AlterByAddColumnContext` behavior.
- Explicit `NULL`/`NOT NULL` constraints and the ClickHouse metadata lookup still take precedence. Only the fallback path (lookup unavailable or failed) changes.

## Known trade-off

If the modified column is part of the ClickHouse table's ORDER BY / primary key **and** the schema lookup is unavailable at the same time, the generated `MODIFY COLUMN ... Nullable(...)` DDL will be rejected by ClickHouse ("Sorting key cannot contain nullable columns"). The old fallback happened to be correct for that narrow case, but silently produced stuck mutations for ordinary nullable columns. The new fallback fails loudly at DDL time instead of corrupting replication silently. The normal path is unaffected: when the metadata lookup succeeds, sorting-key columns still resolve to non-Nullable.

## Tests

- Updated 6 unit test expectations in `MySqlDDLParserListenerImplTest` that pinned the old non-Nullable fallback (unit tests run without a DB connection, so they exercise exactly this path).
- Added 4 new unit tests verifying explicit constraints still win over the new default: `testModifyColumnWithNotNull`, `testModifyColumnWithExplicitNull`, `testChangeColumnWithNotNull`, `testChangeColumnWithExplicitNull`.
- `MySqlDDLParserListenerImplTest`: 99 run, 0 failures.
- `AlterTableModifyColumnIT` (testcontainers, live MySQL + ClickHouse): passes unchanged, confirming the live-connection path is unaffected.

## Files Changed

- `sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImpl.java` (2 branches)
- `sink-connector-lightweight/src/test/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImplTest.java` (6 updated expectations, 4 new tests)

### DIFF
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImpl.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImpl.java
@@ -781,6 +781,10 @@ public class MySqlDDLParserListenerImpl extends MySQLDDLParserBaseListener {
         } else if (tree instanceof MySqlParser.AlterByModifyColumnContext) {
             modifier = Constants.MODIFY_COLUMN;
             modifierWithNull = Constants.MODIFY_COLUMN_NULLABLE;
+            // In MySQL, MODIFY COLUMN without an explicit NULL/NOT NULL constraint
+            // makes the column nullable, so default to Nullable when the current
+            // schema cannot be retrieved from ClickHouse.
+            isNullColumn = true;
         } else if (tree instanceof MySqlParser.AlterByRenameColumnContext) {
             modifier = Constants.RENAME_COLUMN;
             modifierWithNull = Constants.RENAME_COLUMN_NULLABLE;
@@ -789,6 +793,7 @@ public class MySqlDDLParserListenerImpl extends MySQLDDLParserBaseListener {
             isAlterChangeColumn = true;
             modifier = Constants.MODIFY_COLUMN;
             modifierWithNull = Constants.MODIFY_COLUMN_NULLABLE;
+            isNullColumn = true;
         } else if (tree instanceof MySqlParser.AlterByAddIndexContext) {
             modifier = Constants.ADD_INDEX;
         } else {
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImpl.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/MySqlDDLParserListenerImpl.java
@@ -793,6 +793,7 @@ public class MySqlDDLParserListenerImpl extends MySQLDDLParserBaseListener {
             isAlterChangeColumn = true;
             modifier = Constants.MODIFY_COLUMN;
             modifierWithNull = Constants.MODIFY_COLUMN_NULLABLE;
+            // Same MySQL semantics as MODIFY COLUMN above.
             isNullColumn = true;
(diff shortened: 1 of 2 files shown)


clchen-dev:
Hi @subkanthi, could you take a look at this PR when you get a chance? The CI workflows also need maintainer approval to run since this is my first contribution here. Thanks!

clchen-dev:
## 🤖 AI Review (Claude)

> This review was performed by an AI assistant (Claude, Anthropic) at the author's request, including tracing the affected code paths and running the test suite locally.

### Correctness

Traced the full `isNullColumn` data flow through `parseAlterTable` on this branch and confirmed the change behaves exactly as described in the PR summary:

- **Explicit `NULL` / `NOT NULL` constraints still take precedence** — the `NullColumnConstraintContext` branch sets `nullExplicitlySet` and overwrites `isNullColumn` after the new initialization.
- **The ClickHouse metadata lookup still takes precedence** — on the success path all three branches (nullable / column-not-found / non-nullable) explicitly assign `isNullColumn`, so the new default is only observable when the lookup is unavailable (`writer == null`) or throws.
- **The new default matches MySQL semantics** — `MODIFY`/`CHANGE COLUMN` replace the entire column definition, and omitting the constraint makes the column nullable. It is also consistent with the existing `ADD COLUMN` default and with the pre-existing "column not found in metadata → nullable" behavior.
- **CHANGE COLUMN metadata lookup uses the old column name** — correct, since it queries the pre-change state of the column.

### Trade-off

One narrow regression scenario exists in the fallback path: if the modified column is part of the ClickHouse sorting key **and** the metadata lookup is down at the same moment, the generated `Nullable(...)` DDL is rejected by ClickHouse instead of succeeding. That failure is loud and recoverable at DDL time, whereas the old fallback silently produced stuck mutations for ordinary nullable columns — a reasonable trade. The normal (lookup-success) path is unchanged. This is now documented in the PR description.

### Tests

- `MySqlDDLParserListenerImplTest`: **99 run, 0 failures** (verified locally, JDK 17 / Maven 3.9.9). Unit tests run without a DB connection, so they exercise exactly the changed fallback path.
- The explicit-constraint matrix is fully covered: MODIFY/CHANGE × NULL/NOT NULL (`testModifyColumnWithNotNull`, `testModifyColumnWithExplicitNull`, `testChangeColumnWithNotNull`, `testChangeColumnWithExplicitNull`).
- Note: the updated expectations pin some pre-existing whitespace quirks of the nullable code path (double space in `Nullable(Bool)  first`, trailing space before `\n`). These predate this PR — the ADD COLUMN nullable path behaves identically — and are harmless to ClickHouse; cleaning them up would belong in a separate PR.

### Verdict

The fix is correct, minimally scoped, consistent with existing behavior, and well tested. Earlier review nits (comment symmetry on the CHANGE COLUMN branch, the missing CHANGE + explicit NULL test, a formatting nit) were addressed in eb3019e.

PR #1340, opened 2026-07-17
Guard against null JDBC params/settings in BaseDbWriter.createConnection

clchen-dev:
## Summary

`config.getString()` can return `null` for `clickhouse.jdbc.params` and `clickhouse.jdbc.settings`. In the lightweight connector, a YAML config line with an empty value (`clickhouse.jdbc.params:`) parses to null. `createConnection` then throws a `NullPointerException` on `isEmpty()`, which is swallowed by the surrounding catch block, so the caller only sees a null connection with no obvious cause.

We ran into this while setting up TLS connections to ClickHouse via `clickhouse.jdbc.params: "ssl=true,sslmode=none"`, and have been running this fix in production.

## Changes

- Null-guard the `jdbcParams` and `jdbcSettings` checks in `BaseDbWriter.createConnection`.

## Tests

- Added `testSplitJdbcPropertiesWithSsl` covering TLS-style JDBC parameters (`ssl=true,sslmode=none`) in `splitJdbcProperties` — this configuration path previously had no test coverage.
- `BaseDbWriterTest`: 2 run, 0 failures.
- A direct NPE regression test through `createConnection` is not feasible: the method catches all exceptions and returns null, so the NPE is not observable from the outside.

## Files Changed

- `sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java`
- `sink-connector/src/test/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriterTest.java`

### DIFF
--- a/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java
+++ b/sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/BaseDbWriter.java
@@ -248,7 +248,7 @@ public class BaseDbWriter {
         try {
             Properties properties = new Properties();
             properties.setProperty("client_name", clientName);
-            if(!jdbcSettings.isEmpty()) {
+            if(jdbcSettings != null && !jdbcSettings.isEmpty()) {
                 properties.setProperty("custom_settings", jdbcSettings);
             } else {
                 properties.setProperty("custom_settings", "allow_experimental_object_type=1,insert_allow_materialized_columns=1");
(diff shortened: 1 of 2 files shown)


clchen-dev:
Hi @subkanthi, this one is a small null-guard fix, would appreciate a review when you have time. CI is waiting for maintainer approval as well. Thanks!

clchen-dev:
The CI failure is unrelated to this change. The build itself succeeded (Maven builds for Kafka/Lightweight and both Docker images), and the job failed only at the final step of `Build Docker image (Lightweight)`, where `docker buildx build ... --push` tries to push `altinityinfra/clickhouse-sink-connector:<tag>-lt` to Docker Hub:

```
ERROR: failed to push altinityinfra/clickhouse-sink-connector:1340-c56138f6...-lt:
failed to authorize: ... 401 Unauthorized: access token has insufficient scopes
```

Since this PR comes from a fork, GitHub doesn't expose the `DOCKERHUB_*` secrets to the run, so the "Login to Docker Hub" step was skipped (`DOCKERHUB_USERNAME` is empty in the job log) and the push was attempted with an anonymous pull-only token. The Kafka image push step is guarded by `if: env.DOCKERHUB_USERNAME != ''`, but the Lightweight build has `--push` hardcoded in the buildx command in `docker-build.yml`, so this step fails for any fork PR — the same failure can be seen on other fork PRs (e.g. run 29484823727).

Issue #1331, opened 2026-07-12
Sink connector crashes with NullPointerException on cold start if ClickHouse isn't ready (race condition)


BheeshmasenaReddy:
The sink connector crashes permanently with a `NullPointerException` if it attempts to connect to clickHouse before clickHouse (or mysql) is fully ready.

### Steps to reproduce
1. Clone the repo and follow the [QuickStart Guide: Lightweight (MySQL)](https://github.com/Altinity/clickhouse-sink-connector/blob/develop/doc/quickstart.md) 
2. Run a clean start:
Note: I used `clickhouse-sink-connector:2.9.1-lt` image.
```bash
   export CLICKHOUSE_SINK_CONNECTOR_LT_IMAGE=altinity/clickhouse-sink-connector:2.9.1-lt
   docker compose -f docker-compose-mysql.yml up --renew-anon-volumes
```
3. Check docker-clickhouse-sink-connector-lt logs:
```bash
   docker logs docker-clickhouse-sink-connector-lt-1
```

### Expected behavior
After running [Test replication](https://github.com/Altinity/clickhouse-sink-connector/blob/develop/doc/quickstart.md#test-replication) steps from `quickstart.md`, the initial MySQL data (e.g., the test database tables) should be snapshotted into ClickHouse, and subsequent inserts/DDL changes should replicate automatically.

### Actual behavior
On a cold start where ClickHouse isn't ready yet, the `clickhouse-sink-connector-lt` fails DNS resolution / connection resulting in `UnknownHostException: clickhouse: Name or service not known`, then crashes.

<details>
<summary>Log excerpt of clickhouse-sink-connector-lt container</summary>

```
2026-07-11 11:44:29.520 ERROR - HTTP request failed
java.net.UnknownHostException: clickhouse: Name or service not known
	at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method) ~[?:?]
	...
2026-07-11 11:44:29.736 ERROR - Error creating ClickHouse connectionjava.sql.SQLException: clickhouse
2026-07-11 11:44:29.782 INFO  - CREATING DEBEZIUM STORAGE Database: create database if not exists altinity_sink_connector
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.sql.Connection.prepareStatement(String)" because "conn" is null
	at com.altinity.clickhouse.sink.connector.db.DBMetadata.executeSystemQuery(DBMetadata.java:639)
[...]

PR #1348, opened 2026-07-25
Fix NullPointerException on cold start when ClickHouse is not reachable

clchen-dev:
## Summary

#1332 addressed #1331 by gating `clickhouse-sink-connector-lt` on healthy dependencies in the MySQL quickstart compose file. That stops the documented quickstart from triggering the race, but it is a compose-level fix with no code change, so the crash it works around is still present: any deployment where ClickHouse is briefly unreachable — Kubernetes, a ClickHouse restart, a DNS blip, a non-compose deployment, or the PostgreSQL quickstart, which #1332 did not cover — still kills the connector the same way. This PR fixes the crash itself and closes the same ordering gap for PostgreSQL.

`BaseDbWriter.createConnection()` logs the failure and returns `null` when ClickHouse cannot be reached. On a cold start, where the connector comes up before ClickHouse is ready, that null connection is passed down and dereferenced:

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.sql.Connection.prepareStatement(String)" because "conn" is null
	at com.altinity.clickhouse.sink.connector.db.DBMetadata.executeSystemQuery(DBMetadata.java:639)
	at com.altinity.clickhouse.debezium.embedded.cdc.DebeziumJdbcStorageOperations.createDatabaseForDebeziumStorage(DebeziumJdbcStorageOperations.java:70)
	at com.altinity.clickhouse.debezium.embedded.cdc.DebeziumChangeEventCapture.setupDebeziumEventCapture(DebeziumChangeEventCapture.java:198)
	at com.altinity.clickhouse.debezium.embedded.cdc.DebeziumChangeEventCapture.setup(DebeziumChangeEventCapture.java:398)
	at com.altinity.clickhouse.debezium.embedded.ClickHouseDebeziumEmbeddedApplication.start(ClickHouseDebeziumEmbeddedApplication.java:281)
	at com.altinity.clickhouse.debezium.embedded.ClickHouseDebeziumEmbeddedApplication.main(ClickHouseDebeziumEmbeddedApplication.java:188)
```

Two layers of error handling are already in place, but neither engages, because both catch only `SQLException`:

- the retry loop inside `executeSystemQuery()`, which is specifically written to re-acquire a connection from the pool via `initiateNewConnectionIfClosed()`
- the `catch (SQLException e)` around `createDatabaseForDebeziumStorage()` in `setupDebeziumEventCapture()`

A `NullPointerException` is unchecked, so it passes straight through both and escapes `main()`. The replication pipeline dies while the container stays `Up`, which is why the failure gives no indication beyond the log line.

Addresses the crash reported in #1331, which is still open.

## Changes

Four places along that path dereference the missing connection:

- **`DBMetadata.executeSystemQuery()`** — report an unavailable connection as a `SQLException` rather than dereferencing null, so the existing retry-and-reacquire path applies and callers that already handle `SQLException` are no longer bypassed. When no pool can supply a replacement, give up immediately instead of sleeping out the whole retry budget: the loop sleeps `1000 * attempt` ms, which is 45s at the default of 10 retries, and every one of those sleeps is futile when no pool was ever created.
- **`HikariDbSource.initiateNewConnectionIfClosed()`** — the `if (dbSource == null)` block was empty, a comment with no statement, and the data source was dereferenced on the next line. It is reached from the retry path above when no pool exists, which is exactly the cold-start case. Throw a `SQLException` naming the database instead.
- **`DebeziumJdbcStorageOperations.getLatestRecordTimestamp()`** — passed the query result straight into `SimpleDateFormat.parse()`. With no result that throws a `NullPointerException`, which the surrounding `catch (ParseException)` does not catch. Return the method's existing `-1` sentinel instead.
- **`ClickHouseBatchRunnable.getClickHouseConnection()`** — closed a possibly-null connection in its `finally` block, guarded only by `catch (SQLException)`, and cached a failed connection. Because the lookup uses `containsKey()`, which returns true for a key mapped to null, a cached null was then returned for that database until the connector restarted.

For the compose files, extend the startup ordering added in #1332 (7772bbd1) to the PostgreSQL quickstart, which still gated `clickhouse-sink-connector-lt` on container *start* (`depends_on: - clickhouse`) and did not wait for `postgres` at all. That requires a healthcheck on `postgres`, which had none, so add the standard `pg_isready` one. The MySQL compose file is intentionally left untouched, since #1332 already covers it.

Deliberately not changed: startup does not block until ClickHouse becomes reachable. That is a behaviour change worth deciding separately. The scope here is that an unreachable ClickHouse produces actionable `SQLException`s and log lines through the existing handling paths, instead of an unchecked NPE out of `main()`.

## Tests

Added `ClickHouseUnavailableTest` in `sink-connector`, plus one case in the existing `DebeziumJdbcStorageOperationsTest` in the lightweight module. Both are plain unit tests and need no containers.

Each one was verified to be a real regression test by reverting the corresponding fix:

| Test | Result without the fix |
|---|---|
| `testExecuteSystemQueryWithNullConnection` | asserts the call does not sleep out its retry budget: fails at **45,351 ms**, passes at **0.015 ms** |
| `testGetClickHouseVersionWithNullConnection` | same check via the other `executeSystemQuery()` entry point reached from `setupDebeziumEventCapture()` |
| `testInitiateNewConnectionWithoutPool` | `NullPointerException: Cannot invoke "com.zaxxer.hikari.HikariDataSource.getConnection()" because "dbSource" is null` |
| `getLatestRecordTimestamp_returnsSentinelWhenQueryHasNoResult` | `NullPointerException: Cannot invoke "String.length()" because "text" is null` |

- `sink-connector` container-free suite: 69 run, 0 failures.
- `sink-connector-lightweight` container-free suite: 124 run, 0 failures, 1 skipped.
- `mvn -B package install -DskipTests=true`: BUILD SUCCESS for all three modules.
- `docker compose config` renders both changed compose files without error, with the connector gated on `clickhouse` and `postgres` health.

Not covered by tests: the two `ClickHouseBatchRunnable` guards. `getClickHouseConnection()` is private and needs live database credentials, so covering it would require an integration test; those two were verified by inspection only.

One behavioural note worth calling out: no longer caching a failed connection means that while ClickHouse is down, each call retries the connection, bounded by the JDBC connect timeout, instead of returning a cached null immediately. That is the cost of being able to recover once ClickHouse comes back.

## Files Changed

- `sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/DBMetadata.java`
- `sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/db/HikariDbSource.java`
- `sink-connector/src/main/java/com/altinity/clickhouse/sink/connector/executor/ClickHouseBatchRunnable.java`
- `sink-connector/src/test/java/com/altinity/clickhouse/sink/connector/db/ClickHouseUnavailableTest.java` (new)
- `sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperations.java`
- `sink-connector-lightweight/src/test/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperationsTest.java`
- `sink-connector-lightweight/docker/docker-compose-postgres.yml`
- `sink-connector-lightweight/docker/postgres-service.yml`

### DIFF
--- a/sink-connector-lightweight/docker/docker-compose-postgres.yml
+++ b/sink-connector-lightweight/docker/docker-compose-postgres.yml
@@ -31,7 +31,10 @@ services:
       file: clickhouse-sink-connector-lt-service.yml
       service: clickhouse-sink-connector-lt
     depends_on:
-      - clickhouse
+      clickhouse:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
     extra_hosts:
       - "host.docker.internal:host-gateway"
     environment:
--- a/sink-connector-lightweight/docker/postgres-service.yml
+++ b/sink-connector-lightweight/docker/postgres-service.yml
@@ -4,6 +4,11 @@ services:
   postgres:
     image: debezium/postgres:15-alpine
     restart: always
+    healthcheck:
+      test: [ 'CMD-SHELL', 'pg_isready -U root -d public' ]
+      interval: 10s
+      timeout: 5s
+      retries: 6
     ports:
       - "5432:5432"
     environment:
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperations.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperations.java
@@ -304,6 +304,14 @@ public class DebeziumJdbcStorageOperations {
             throws SQLException {
         long result = -1;
         String latestRecordTs = this.debeziumOffsetStorage.getDebeziumLatestRecordTimestamp(props, conn);
+        if (latestRecordTs == null || latestRecordTs.isEmpty()) {
+            // The query returned no result, e.g. because ClickHouse was not
+            // reachable. Without this guard SimpleDateFormat.parse() below
+            // throws a NullPointerException, which the ParseException handler
+            // does not catch.
+            log.error("Latest record timestamp is not available");
+            return result;
+        }
         // Convert date string from "yyyy-MM-dd HH:mm:ss" format to milliseconds.
         SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
         Date date = null;
--- a/sink-connector-lightweight/src/test/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperationsTest.java
+++ b/sink-connector-lightweight/src/test/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumJdbcStorageOperationsTest.java
@@ -47,4 +47,21 @@ public class DebeziumJdbcStorageOperationsTest {
 
         assertDoesNotThrow(() -> ops.createSchemaHistoryTable(null, props));
     }
+
+    @Test
+    @DisplayName("getLatestRecordTimestamp returns -1 without NPE when the query has no result")
+    void getLatestRecordTimestamp_returnsSentinelWhenQueryHasNoResult() {
+        DebeziumJdbcStorageOperations ops = new DebeziumJdbcStorageOperations();
+        Properties props = new Properties();
+        String offsetTableKey = JdbcOffsetBackingStoreConfig.OFFSET_STORAGE_PREFIX +
+                JdbcOffsetBackingStoreConfig.PROP_TABLE_NAME.name();
+        props.setProperty(offsetTableKey, "altinity_sink_connector.replica_source_info");
+
+        // A null connection makes the underlying query return no result. That
+        // reached SimpleDateFormat.parse(null) and threw a NullPointerException,
+        // which the surrounding ParseException handler does not catch.
+        long result = assertDoesNotThrow(() -> ops.getLatestRecordTimestamp(null, props));
+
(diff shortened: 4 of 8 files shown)


clchen-dev:
### AI-assisted review

I ran an AI-assisted review over my own first version before pushing. It found two real problems, both fixed in `520d5b13`:

1. **The NPE was relocated, not removed.** `getLatestRecordTimestamp()` passed the now-`null` query result into `SimpleDateFormat.parse()`, which throws an NPE that the surrounding `catch (ParseException)` does not catch.

2. **It slept through the full retry budget.** A null connection routed into the existing retry loop took a measured **45,351 ms** to give up, where the previous behaviour failed in milliseconds. The loop can only recover a null connection from the pool, which does not exist on a cold start, so every one of those sleeps was futile. It now fails fast in that case, and still retries normally when a pool does exist (e.g. a ClickHouse restart under a running connector).

It also surfaced two pre-existing dereferences of the same null connection in `ClickHouseBatchRunnable.getClickHouseConnection()`, now guarded.

Each fix has a regression test that was verified by reverting the fix. Remaining coverage gaps are listed in the description. Happy to split the `ClickHouseBatchRunnable` guards into a separate PR if you would rather keep this one to the startup path.

clchen-dev:
Hi @subkanthi, please help take a look when you have time, thanks!

Issue #1346, opened 2026-07-23
Delete re-insert version mismatch


subkanthi:
bulk DELETE all rows, then re-INSERT them (e.g., a nightly refresh) — causes permanent data loss on the ClickHouse DR replica running connector version 2.9.1, while 2.8.0 handles it correctly.

subkanthi:
<img width="1676" height="1146" alt="Image" src="https://github.com/user-attachments/assets/2d530c88-3ffb-4ce9-bf87-a9b82c7a164e" />


Keep the existing ts * 1_000_000 + counter version scheme (so snapshot vs. streaming magnitudes stay comparable and nothing regresses), but anchor it to the source commit timestamp (source.ts_ms), which is identical every time Debezium re-delivers an event and correctly orders DELETE (earlier commit) before its later re-INSERT.

Issue #1140, opened 2025-09-17
Multiple add column are broken with algorithm


aadant:
```
CREATE TABLE `test_lot` (
  `lot_id` bigint unsigned NOT NULL,
  `trade_date` date NOT NULL,
  PRIMARY KEY (`lot_id`,`trade_date`)
) ENGINE=InnoDB 
  
 ALTER TABLE test_lot ADD COLUMN event_ref_type_id INTEGER, algorithm=instant, ADD COLUMN event_ref_id BIGINT, algorithm=instant;
```

subkanthi:
<img width="687" height="1016" alt="Image" src="https://github.com/user-attachments/assets/a16a1f91-7552-40d6-8e5b-d46248ea3070" />

subkanthi:
Created issue in Debezium https://issues.redhat.com/browse/DBZ-9704 and moving to 2.9.0

PR #1353, opened 2026-07-31
DDL/DML data-integrity fixes: schema-cache races, _version collisions, DDL translation

minguyen9988:
> **WIP: DO NOT MERGE** — opened for upstream review and discussion. Please do not merge without maintainer review; happy to split this into per-area PRs if that is easier to review.

Builds on `2.10.0`. These are fixes for silent data loss and replication stalls found while running the connector against a high-volume MySQL → ClickHouse workload. Every fix has a regression test that was verified to **fail against the unfixed code** and pass against the fix.

## 1. Schema-cache / DDL race — silent data loss (#1222)

The batch thread builds its `INSERT` from the **cached ClickHouse column map**, not from the source event fields. When a DDL adds a column and a batch thread inserts using a cache that predates the `ALTER`, the new column is omitted and ClickHouse stores the `DEFAULT` (NULL/0) instead of the source value. No error is raised — the row lands with a wrong value.

The existing DDL waiter is necessary but not sufficient: it makes the *DDL thread* wait for `system.columns` visibility, but it does not stop *other batch threads* inserting against the stale cache in the meantime.

**Fix:** a per-table replication freeze held until (a) the destination schema is visible, (b) the cache is invalidated, and (c) source-vs-destination integrity is confirmed — plus a cache TTL that self-heals a DDL that was missed entirely.

## 2. `_version` collisions — ReplacingMergeTree row loss

Three distinct defects, each of which let two records receive the **same** `_version`, so RMT silently collapsed one of them.

| # | Defect | Reproduction |
|---|---|---|
| 1 | Counter incremented with `incrementAndGet()` but the value discarded and re-read with a separate `get()`. Two threads interleaving between the increment and the read observe the same value. | 406 duplicate versions out of 6400 |
| 2 | The time-gap reset assigned a **constant**, and the source-timestamp anchor was a per-call local re-seeded from `record[0]` on every call. Every batch spanning a gap reset to the same constant — a *deterministic* collision, not a lost update, so locking alone did not fix it. | 8 concurrent batches → 1 distinct version |
| 3 | `calculateVersion()` is a no-op when a record carries no GTID, binlog position, sequence number or LSN, leaving the `-1` sentinel in place — and the bind site wrote it unconditionally. | see below |

Defect 3 is the most severe of the three. `_version` is `UInt64` in ClickHouse and the value is bound with `setLong`, so `-1` is stored as **18446744073709551615** — the maximum representable version. Verified against a live ClickHouse:

```sql
SELECT reinterpretAsUInt64(toInt64(-1));  -- 18446744073709551615
```

A row carrying that value **can never be superseded**: the ReplacingMergeTree freezes on it and every later `UPDATE` and `DELETE` for that key is silently discarded on merge. The bind now throws instead, naming the topic, offset and the value that would have been written.

## 3. MySQL → ClickHouse DDL translation

- **FLOAT/DOUBLE precision loss.** Debezium widens MySQL `FLOAT` to a `FLOAT64` Kafka schema, so `FLOAT` and `DOUBLE` are indistinguishable by schema type and a single mapping *must* be wrong for one of them: mapping to `Float32` truncates every `DOUBLE` from ~15 significant digits to ~7 **on every replicated row**; mapping to `Float64` doubles the width of real `FLOAT` columns. Now decided from the resolved source JDBC type (`FLOAT`/`FLOAT4` → `Float32`, `DOUBLE`/`FLOAT8`/`REAL` → `Float64`).
- **`CHANGE COLUMN` emitted two databases from one statement** — the `MODIFY` half targeted the destination database while the `RENAME` half kept the source-qualified name, so with `database.override.map` set the rename hit the wrong table or failed.
- **Dangling separator after an unsupported clause.** `ALGORITHM=` / `LOCK=` emit nothing but left their separator behind, producing a trailing or doubled comma — a ClickHouse syntax error that rejects the whole `ALTER` and **stalls the DDL stream**.
- **`DROP` now always emits `IF EXISTS`.** A `DROP` replayed after an offset rewind, or for a table that was never replicated, raised `UNKNOWN_TABLE` and stalled the stream.

## 4. Safety and correctness

- `disable.drop.truncate` is now enforced on the CDC data path, and can actually be set `false` in the Helm chart. Sprig's `default` treats boolean `false` as empty, so an explicit `false` was silently read back as `true` — an operator *deliberately disabling* the guard was being ignored. Verified across all 8 render combinations of the flag.
- Offset-storage deletes bind their parameters.
- A swallowed schema-cache exception now logs.
- Null-valued fields are no longer dropped from `UPDATE`s.
- GTID parsing handles multi-range and multi-source forms.

## Tests

New compatibility and regression suites, all pure unit tests (no container required) so they run on every build:

- **`VersionCompatibilityTest`** (14 tests) — pins the `_version` data-format contract, because a change to it is irreversible for a live table:
  - *redelivery stability*: a replayed event must keep its version, or a replayed `DELETE` outranks the re-`INSERT` (#1346);
  - *commit ordering* across binlog file rollovers;
  - *uniqueness* across 512 positions and 512 seconds;
  - *unsigned ordering across the 2038 sign-bit boundary* — at `2038-01-19T03:14:08Z` the commit second reaches 2^31 and sets the sign bit, so the Java `long` goes negative while the stored `UInt64` keeps increasing correctly. Ordering holds in the database, but only as long as no Java-side code compares versions with a signed operator; this test pins that requirement;
  - *magnitude envelope*, so a future scheme change that silently makes a downgrade impossible fails here rather than in production.
- **`DDLTranslationRegressionTest`** (13 tests) — pins the exact emitted SQL for each translation defect above, plus forward-compatibility cases where unknown MySQL syntax must degrade to valid SQL rather than stall the stream.
- **`VersionSentinelGuardTest`** (3 tests) — covers the `-1` sentinel rejection; confirmed to fail against the unfixed bind.

```
sink-connector              392 run / 0 failures / 3 errors
sink-connector-lightweight  199 run / 0 failures / 2 errors
```

All 5 errors are testcontainer Docker-environment failures (image pull unavailable on the build host) and were **reproduced identically on an unmodified `2.10.0` checkout** — environmental, not introduced by this change.

## Notes for reviewers

- Based on `3bff01c5` (the `2.10.0` tip at the time of writing).
- The `_version` formula itself is **unchanged** — only the paths that could produce a duplicate or a sentinel value are fixed. Changing the formula would be irreversible for existing tables, which is what `VersionCompatibilityTest` now guards.
- Happy to split this into per-area PRs (lightweight / core connector / Helm) if that is easier to review.

## Update: rebased onto the current `2.10.0` tip (`13e131df`)

Upstream reverted #1347 in #1352 while this PR was open, which removed the binlog-position branch from `calculateVersion()`. Issue #1346 itself is still open.

**This PR adopts that revert rather than re-introducing the reverted formula.** The sequence-number path now follows what was described on #1346: retain the `ts * 1_000_000 + counter` encoding, but anchor it to the **source commit timestamp** (`source.ts_ms`) rather than the Debezium processing timestamp, which changes on every redelivery and is what allowed a re-delivered `DELETE` to outrank a live re-`INSERT`. The concurrency fixes on that path are unchanged: the counter and its time anchor move together under one lock, and the anchor is instance state rather than a per-call local.

`VersionCompatibilityTest` was rewritten accordingly - it now tests the version sources `calculateVersion()` actually consults (GTID, snowflake, sequence number, LSN) instead of the removed branch, and gains a **version-source precedence** group. That precedence (`GTID` > `sequenceNumber` > `LSN`) is itself part of the data format: a build that resolved a given row from a different source than the build before it would write a competing value range into the same ReplacingMergeTree.

Test totals on the current tip:

```
sink-connector              398 run / 0 failures / 3 errors
sink-connector-lightweight  199 run / 0 failures / 2 errors
```

## Update 2: `_version` restored to the 2.8.0 formula

**2.10.0's `_version` calculation is not backward compatible with 2.8.0.** This PR now reverts it. The emitted value is bit-for-bit the 2.8.0 formula:

```
_version = debezium_ts_ms * 1_000_000 + counter
```

with the counter reset to `SEQUENCE_START` when the Debezium clock advances more than one second past the anchor (`diff > 1`), and incremented otherwise.

`_version` decides which physical row survives a ReplacingMergeTree merge, which makes the formula a **data format**, not an implementation detail. A table that already holds rows written by 2.8.0 and then receives rows produced by a different formula ends up with two incomparable magnitudes in one column: whichever scheme yields the larger number wins every merge regardless of which change actually happened later, and the rows already written cannot be undone by downgrading. It is a one-way door.

An earlier revision of this branch anchored the sequence on the **source** commit timestamp (`source.ts_ms`) instead. That is a strictly better ordering property — it is stable across at-least-once redelivery, which is the root of the still-open #1346 — but it shifts every emitted value, so it cannot be adopted on a branch that must stay readable alongside 2.8.0 data. It is reverted here and left for a separate discussion with a migration story.

**What is not reverted is the concurrency defect.** The shared counter was read-modify-written without a lock in two windows: the reset branch is a check-and-act across the anchor and the counter, and the increment branch called `incrementAndGet()` and then re-read the counter with a separate `get()`. Two batch threads interleaving in either window observe the same counter and hand two distinct records an identical `_version`, which the ReplacingMergeTree silently collapses into one row. The batch-local anchor also shadowed the shared instance field, letting two batches independently reset to `SEQUENCE_START`. Both are fixed by assigning under one lock, with the value taken from inside the critical section. **The emitted values are unchanged.**

The live batch consumer and `addVersion()` now share one helper, so the two paths cannot drift into different schemes.

`Version280CompatibilityTest` (6 tests) pins the formula against an independent reimplementation taken from the `2.8.0` tag — including the `diff > 1` reset boundary and the `debezium_ts_ms` anchor — plus an 8-thread x 400-record concurrency test. Verified to fail (4 of 6) against a deliberately altered formula.

```
sink-connector              398 run / 0 failures / 3 errors
sink-connector-lightweight  205 run / 0 failures / 2 errors
```


minguyen9988:
Rebuilt against the current `2.10.0` tip (`8ec596af`) — the branch was 20+ commits behind and `CONFLICTING`; it is now `MERGEABLE` (merge commit `ed5dbc4a`).

Ten files conflicted. Three of them needed code that is in **neither** parent, because taking either side verbatim would have silently dropped a fix. Flagging those explicitly:

**1. `PreparedStatementExecutor` — the one worth reviewing.** #1382 deliberately deleted the end-of-batch TRUNCATE block and now truncates at the record's binlog position. This branch's entire contribution in that file was the `disable.drop.truncate` guard on that path. Accepting `2.10.0` verbatim would have removed the guard and left the CDC data path applying every TRUNCATE unconditionally again — a silent regression, since the setting would still be accepted and would still do nothing. The guard is ported onto the new in-position call site instead, and this branch's older "truncate before inserts" hoist is dropped, as #1382's placement supersedes it.

**2. `DebeziumChangeEventCapture` — the two sides fix different halves of the same `_version` defect.**
- `2.10.0` fixes *which* timestamp is used: the source commit ts, so a re-delivered DELETE keeps its original `_version` (#1346), plus the `SEQUENCE_START_INITIAL` post-resume seed (#1378).
- This branch fixes *how* the counter is assigned: the inline read-modify-write is a check-and-act race that hands two records an identical `_version`, which ReplacingMergeTree then collapses.

Taking either alone reintroduces the other's bug, so the source-ts anchoring and the 500m seed now live **inside** the locked `nextSequenceNumber` helper, and both call sites (batch loop and `addVersion`) go through it. The batch-level `seedSequenceAnchor` is removed: with per-record source-timestamp anchoring there is no single correct timestamp to anchor a whole batch on, and a per-batch anchor was itself a race. The emitted formula (`ts_ms * 1_000_000 + counter`) is unchanged from 2.8.0 on every path — that is a one-way door and was not touched.

**3. `DBMetadata.executeSystemQuery`.** `2.10.0` switched to `execute()` so DDL/DML returning no ResultSet works with strict drivers (clickhouse-jdbc >= 0.9.x rejects `executeQuery()` for those); this branch used try-with-resources because the old code leaked a `PreparedStatement` on every retry. Both are needed — merged as `execute()` inside try-with-resources.

The rest, briefly: `ClickHouseDataTypeMapper` keeps 2.10.0's `setGeoValue` (a bare `ps.setObject` breaks under the V2 driver) plus this branch's WKB parse-failure warning; `HikariDbSource` takes 2.10.0's system-pool fallback and closed-pool drop, with this branch's error log re-added so the operator can see which database never got a pool; `ClickHouseBatchRunnable`/`Writer` take `isUnusable()`, which subsumes this branch's lazy-open; `MySqlDDLParserListenerImpl` takes 2.10.0's version (it also covers `AlterByLockContext`) while keeping this branch's whitespace-tolerant helpers, still called elsewhere in the walk; `ClickHouseBatchExecutor` was a docs-only conflict on the same `volatile` field.

One merge artifact worth noting: in `DBMetadataTest` the conflict was confined to the import block, but resolving only the imports left the class without `@Testcontainers`, the `@Container` field and the `@AfterAll` cleanup — they sit on the 2.10.0 side just below the conflict region, and a test added by this branch references them. Restored all four.

**Tests**

```
                            Tests run   Failures   Errors
sink-connector                    476          0       18
```

All 18 errors are testcontainers `Could not find a valid Docker environment` and reproduce identically on the untouched baseline — environmental on my build host, not merge-introduced. Both modules compile clean. CI is running now and is the authoritative check for the container-backed tests I could not execute here.

minguyen9988:
Closing this. Its content has since landed piecemeal upstream, and the one defect still live has been split into its own PR.

I re-tested every claim in this PR against the current `2.10.0` head (`78b868c3`) on the published image `1304-364c719120e69c99a389460b46c02e2f43b0f00f-lt`, MySQL 8.0.36 ROW/FULL/GTID -> ClickHouse 24.8.14.10547, rather than reading the changelog:

| # | claim | status on current head | evidence |
|---|---|---|---|
| 1 | schema-cache / DDL race (#1222) | fixed elsewhere | #1321 + #1384; `DDLSchemaChangeWaiter` + `CacheInvalidationManager` present |
| 2 | `_version` collisions (3 defects) | fixed elsewhere | 8 writers x 80 rows: 640 rows, **640 distinct `_version`, 0 duplicates, 0 sentinels**, none > 2^63 |
| 3a | DOUBLE precision loss | fixed by #1386 | DOUBLE / DOUBLE PRECISION / FLOAT8 all exact. **REAL still wrong -> new PR** |
| 3b | CHANGE COLUMN emitted two databases | fixed | emits `MODIFY COLUMN` + `RENAME COLUMN` correctly |
| 3c | ALGORITHM / LOCK dangling separator | fixed (#1140) | multi-op `ALTER ... ALGORITHM=INSTANT` and `LOCK=NONE` both applied |
| 3d | DROP without IF EXISTS | still absent | emits bare `DROP TABLE`; no stall observed — hardening, not corruption |
| 4 | null field dropped from UPDATE | fixed | `SET b=NULL` -> NULL on both sides |
| 4 | `disable.drop.truncate` on the CDC data path | still DDL-path only | TRUNCATE propagated 3 -> 0 rows with the guard unset — hardening, not corruption |

The one live data defect, **REAL declared `Float64` while carrying float32 precision**, is now #1397 with a failing-first test. Root cause turned out to be upstream in Debezium (`Types.REAL` -> float32 schema), which I verified by building a destination-aware bind and measuring that it changed nothing.

Rebasing 118 files across 12 merged commits to salvage one narrow fix is not a good trade, so this is better closed than carried. The two remaining hardening items (`DROP IF EXISTS`, CDC-path TRUNCATE guard) are real but non-corrupting — I'm happy to open small separate PRs for them if you want them.

Also noted while testing and not yet filed: a MySQL `FLOAT4` column is dropped from the INSERT entirely and lands NULL, reproduced on the unmodified published image.

PR #1362, opened 2026-08-05
[split 7/22] deduplicator: fix eviction bug that never added keys to the FIFO pool, add concurrency test

minguyen9988:
DeDuplicator.updateDedupePool() iterated over ALL topics' pools and attempted matchingQueue.remove(key) on an element already removeFirst()d - while never adding the new key to the current topic's queue. Eviction therefore never tracked insertions and the dedup map could grow without bound. Now: key appended to the per-topic FIFO, oldest entries evicted (from both queue and map) when maxPoolSize is exceeded. DeDuplicatorConcurrencyTest exercises the pool under parallel writers; DeDuplicatorTest expanded.


minguyen9988:

PR #1365, opened 2026-08-05
[split 10/22] model: ClickHouseStruct hardening and 2.8.0 version-format compatibility pinning tests

minguyen9988:
- ClickHouseStruct: defensive null handling in Kafka-metadata extraction, accessor cleanup. BlockMetaData exposes min/max source timestamps and record counts for observability.
- VersionCompatibilityTest (new): pins the on-the-wire _version semantics to the 2.8.0 format - debezium_ts_ms * 1_000_000 + counter with diff > 1 reset - so a drop-in upgrade cannot silently change which row survives a ReplacingMergeTree merge.
- RoutedBatchTest rewritten against real behavior.


minguyen9988:

PR #1366, opened 2026-08-05
[split 11/22] common: Utils null-safe topic-to-table mapping, isValidTable(), SnowFlakeId determinism tests

minguyen9988:
- Utils.parseTopicToTableMap/parseSourceToDestinationDatabaseMap: null/empty input returns an empty map instead of NPE.
- New Utils.isValidTable() - ClickHouse identifier validation (length, first char, allowed charset); covered by UtilsAdditionalTest. Not yet called by shipped code - wiring arrives with the executor work.
- SnowFlakeId: doc + overflow-behavior fix; SnowFlakeIdTest expanded to pin bit layout and monotonicity.
- Logging string-format fix.


minguyen9988:

PR #1367, opened 2026-08-05
[split 12/22] history: quote database identifier in BinLogHistory DDL, null-safe host columns

minguyen9988:
BinLogHistory emitted CREATE TABLE with the database name unquoted (breaks on hyphenated database names) and could return null for host columns (NPE at insert bind time). Identifiers are now backtick-quoted and host columns null-coalesced. BinLogHistoryTest expanded.


minguyen9988:

PR #1368, opened 2026-08-05
[split 13/22] db: DBMetadata engine detection robustness, BaseDbWriter/Hikari connection lifecycle

minguyen9988:
- DBMetadata: engine detection (getTableEngine / RMT version-column parsing) hardened against unexpected system.tables output; result-set and statement lifecycles closed via try-with-resources. DBMetadataTest rewritten accordingly.
- BaseDbWriter: connection-creation failures logged with target-URL context and rethrown (previously half-swallowed); shared SYSTEM_DB constant.
- HikariDbSource: pool lookup for a database that was never initialized returns a clear error instead of NPE (HikariDbSourceTest).
- ErrorLogger format fix.


minguyen9988:

PR #1370, opened 2026-08-05
[split 15/22] config: disable.drop.truncate + ddl.schema.change.* settings, validator NPE fixes

minguyen9988:
Stacked PR: based on the previous PR in the series (`2.10.0`); this PR's own diff is only its listed files. Merge the series in order; after the predecessor merges, retarget this PR to `2.10.0`.

First of the ordered core series - later core PRs build on these config keys.

- New config keys: disable.drop.truncate (guards replicated DROP/TRUNCATE), ddl.schema.change.timeout.ms / ddl.schema.change.poll.interval.ms (used by the DDL schema-change waiter in a follow-up PR).
- ClickHouseSinkConnectorConfig: parse-before-guard ordering so boolean guards read the parsed value, not the raw default.
- DatabaseOverrideValidator / TopicToTableValidator / ColumnOverrides / SchemaOverrideConfig: null-input NPE fixes, dead code removed.
- ClickHouseSinkTask / ClickHouseSinkConnector: minor lifecycle logging fixes.

Merge order: this PR first; then db-ops, ddl-parser, core classes, batch, executor, lw-engine.


minguyen9988:

PR #1355, opened 2026-08-02
Fix delete/re-insert flow , added integration test for backwards comp…

subkanthi:
…atibility with 2.8.0

closes: #1346 
extract the delete/re-insert fix from https://github.com/Altinity/clickhouse-sink-connector/pull/1353

PR #1378, opened 2026-08-06
Fix #1346: redelivery-stable _version anchored to source commit timestamp - keeps the 2.8.0 formula (upgrade- and rollback-safe)

### LINKED ISSUE #1346: Delete re-insert version mismatch

minguyen9988:
Closes #1346.

## Problem

Bulk DELETE of all rows followed by re-INSERT (e.g. a nightly refresh) causes permanent data loss on the ClickHouse replica: after an offset rewind, Debezium re-delivers the DELETE with a **fresh processing timestamp**, so it receives a **higher `_version`** than the later re-INSERT, and ReplacingMergeTree keeps the row permanently `is_deleted=1`.

## Fix - redelivery-stable versions **without changing the version formula**

Anchor the version sequence to the **source commit timestamp** (`source.ts_ms`) instead of the envelope/processing timestamp. The source timestamp is identical on every re-delivery, so a re-delivered DELETE keeps its original (lower) version and can never out-rank the later re-INSERT.

Unlike #1347 (reverted in #1352) and #1355, the emitted formula is **unchanged from 2.8.0**: `ts_ms * 1_000_000 + counter`.

- **Upgrade-safe**: new values live in the same numeric domain as existing 2.8.0/2.9.x/2.10.0-written rows; later commits still rank higher.
- **Rollback-safe**: because the domain is unchanged, downgrading the binary later remains possible - no un-supersedable version range is introduced (with `(sourceSec << 32) | pos`, values jump to ~7.6e18 vs ~1.78e18, and a rolled-back binary could never supersede them until the year ~2213).

### Binary log rotation

The intra-second counter is keyed **exclusively on the source commit clock** - never on the binlog file name or position - so it is preserved across binary log rotations: two commits in the same second on either side of a rotation keep incrementing the same counter and cannot collide or invert. The anchor is global (survives batch boundaries) and never moves backward, which also closes the race where a re-delivered older event re-armed the counter reset and produced duplicate `_version` values.

### Snapshot records

Also fixes snapshot-marker parsing: Debezium emits an enum string (`"true"`, `"first"`, `"last"`, `"incremental"`, ...), not a boolean - `Boolean.parseBoolean` misclassified `"first"`/`"last"` as streaming records. Snapshot records keep the historical processing-time anchor (their `source.ts_ms` is the snapshot read time with different clock semantics, and snapshots are re-read on rewind, not re-delivered, so #1346 does not apply to them).

## Tests

`SourceTsVersionAnchorTest` (6 tests) pins:
1. the emitted `_version` stays bit-compatible with the 2.8.0 domain,
2. the #1346 scenario - a re-delivered DELETE cannot out-rank the later re-INSERT,
3. the intra-second counter is kept across binlog rotation,
4. the counter resets only when the source clock advances >1s (2.8.0 semantics),
5. the anchor never moves backward on redelivered older events,
6. records without `source.ts_ms` fall back to the processing timestamp (no version-0).

**All 6 tests fail against the previous logic and pass with the fix** (verified by reverting the fix locally). Existing `DebeziumChangeEventCaptureTest` passes unchanged.

### DIFF
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
@@ -164,8 +164,20 @@ public class DebeziumChangeEventCapture {
     public static final long SEQUENCE_START = 1000000000;
 
     /**
-    * Initial starting sequence number.
-    */
+     * Initial starting sequence number, used ONLY for the first batch after the
+     * connector starts/resumes (500 million, half of {@link #SEQUENCE_START}).
+     *
+     * <p>On a resume, Debezium re-publishes every event after the last committed
+     * offset — including events that were already delivered and written before the
+     * shutdown/crash with counters in the {@code SEQUENCE_START} (1&nbsp;billion) range.
+     * Seeding the post-resume counter at 500 million guarantees every re-published
+     * event in the same source second receives a strictly LOWER {@code _version} than
+     * any pre-restart write of that second, so a re-published duplicate can never
+     * supersede a row that was already correctly written. The first source-clock
+     * advance of more than one second resets the counter to {@code SEQUENCE_START},
+     * returning to the normal domain. Values stay inside the 2.8.0 numeric domain
+     * ({@code ts_ms * 1_000_000 + counter}) in both phases.</p>
+     */
     public static final long SEQUENCE_START_INITIAL = 500000000;
     
     /**
@@ -272,8 +284,14 @@ public class DebeziumChangeEventCapture {
                         // boundaries) and never moves backward, so re-delivered events with
                         // older source timestamps cannot re-arm the counter reset (the
                         // duplicate-_version race).
+                        //
+                        // First record after start/resume: seed the counter at
+                        // SEQUENCE_START_INITIAL (500m) so events re-published from the last
+                        // committed offset rank strictly below any pre-restart write of the
+                        // same source second (which carried counters in the 1000m range).
                         if (sequenceAnchorTs == 0L) {
                             sequenceAnchorTs = recordTs;
+                            sequenceNumber = SEQUENCE_START_INITIAL;
                         }
                         int diff = (int) ((recordTs - sequenceAnchorTs) / 1000);
                         if (diff > 1) {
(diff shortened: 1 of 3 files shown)


minguyen9988:
Pushed `8222636` addressing @aadant's directives, each with exact, provable testing (all in `SourceTsVersionAnchorTest`, now 9 tests):

**1. The intra-second counter is kept even if the binary log rotates.**
The counter is keyed exclusively on the source commit clock — the binlog file name and position play no role. Proven by `counterKeptAcrossBinlogRotation` (same-second batches on either side of a rotation keep incrementing: `…+4`, never reset) and additionally inside a re-publication scenario where the rotation falls between the re-published DELETE and re-INSERT (`republicationPreservesOrderAcrossRotation`).

**2. The first time you resume, the counter is 500m.**
Restores the historically intended initial-seed semantics (removed upstream in 686cfea7): on the first batch after start/resume (`sequenceAnchorTs == 0`), the counter is seeded at `SEQUENCE_START_INITIAL` (500,000,000) instead of the steady-state `SEQUENCE_START` (1,000,000,000). Debezium re-publishes every event after the last committed offset on resume; those events were already written pre-restart with counters in the 1000m range, so the 500m seed guarantees every re-published duplicate ranks strictly *below* its original write and can never supersede it. The first >1s source-clock advance resets to `SEQUENCE_START`, returning to the normal domain. Both counter sites (streaming loop + `addVersion`) are seeded. Proven by `resumeSeedsCounterAtInitial` and `initialSeedEscapesToNormalDomainAfterOneSecond`.

**3. Edge cases including re-publication.**
`republicationPreservesOrderAcrossRotation`: crash + resume replays DELETE and re-INSERT across a binlog rotation — DELETE < re-INSERT ordering is preserved, the replayed DELETE ranks below the ORIGINAL re-INSERT already in ClickHouse (the #1346 case cannot return on resume), and no replayed version out-ranks its original write. Plus the existing pins: `redeliveredDeleteCannotOutrankReinsert`, `anchorNeverMovesBackward`, `counterResetsOnlyOnSourceClockAdvance`, `staysInTwoEightZeroDomain` (both the 500m and 1000m phases stay inside the 2.8.0 `ts_ms*1e6+counter` domain), `fallsBackToProcessingTimestamp`.

**Proof of test sensitivity (negative control):** with the seed reverted (previous head `073a2864`), 6 of the 9 tests fail; with it applied, 9/9 pass.
**Zero regressions:** the full offline suite (63 classes) was run on both this commit and the previous head — the failing-class sets are IDENTICAL (47 pre-existing Docker-unavailable/environment classes). `Version280CompatibilityTest` (6/6), `DebeziumChangeEventCaptureTest` (13/13) and `SequenceNumberRaceTest` (7/7) all pass.

PR #1394, opened 2026-08-23
fix: keyless source tables collapse to a single row in ClickHouse

minguyen9988:
**Requires human review before merging.

Reviewed by a 3-model cross-provider adversarial judge panel (Gemini 3.1 Pro / Grok 4.2 Reasoning / GPT-5.5): 3/3 APPROVE, round 1, zero blocking findings.

Full detail follows, as committed:

fix: keyless source tables collapse to a single row in ClickHouse

A MySQL table that declares neither a PRIMARY KEY nor a UNIQUE key was
created in ClickHouse as `ReplacingMergeTree(...) ORDER BY tuple()`. With
an empty sorting key every row compares equal, so ReplacingMergeTree keeps
exactly ONE row for the entire table.

Measured on MySQL 8.0.36 (ROW/FULL/GTID) -> ClickHouse 24.8.14, MySQL row
counts on the left:

    scenario                        mysql   clickhouse
    5 distinct rows                     5            1   total loss
    3 rows, UPDATE one of them          3            1   total loss
    3 rows, DELETE one of them          2            0   table emptied
    grows from 1 row to 2 rows          2            1   loss on growth

The canonical instance is Alembic's `alembic_version`, which holds exactly
one row -- so the collapse is invisible there and appears the moment any
keyless table holds two. The defect is a property of the table SHAPE, not
of Alembic. On one staging cluster 38 tables have an empty sorting key.

Fix, in two halves.

DDL: when a table declares no PRIMARY KEY and no UNIQUE key, order by every
column. The full row becomes its own identity, which is exactly MySQL's
semantics for a keyless table -- rows are distinguished by their values.
`PRIMARY KEY tuple()` is emitted alongside so the in-memory sparse index
stays empty despite the wide sorting key, and `allow_nullable_key=1`
because such a key necessarily spans the table's nullable columns
(ClickHouse otherwise rejects it with Code: 44 ILLEGAL_COLUMN). Columns are
taken in declaration order so the key is reproducible across connectors
replicating the same source; generated columns are excluded, being a pure
function of the stored columns and absent from the CDC payload.

Writer: ReplacingMergeTree deduplicates by sorting key, so an UPDATE that
changes a sorting-key column writes the new row at a different key and
leaves the pre-update row in place forever. The before-image is now
emitted as a tombstone when, and only when, the key actually changes.

That second half is not specific to keyless tables. A keyed table orphans
the old row in exactly the same way -- `ORDER BY (id)`, `UPDATE id=1->9`
left ClickHouse with 3 rows against MySQL's 2. It is merely guaranteed to
happen on a keyless table, whose sorting key is every column.

The tombstone is conditional for a reason: for a same-key UPDATE the delete
marker and the new row would land on the same sorting key, and
ReplacingMergeTree breaks a version tie by insertion order. Measured with
the tombstone written second, the live row is permanently lost (FINAL
returns 0 rows) and stays lost across merges.

End-to-end against live MySQL and ClickHouse, MySQL as the reference:

    scenario                                     mysql   ch   result
    alembic 1 row                                    1    1   PASS
    alembic migration bump (UPDATE)                  1    1   PASS
    five distinct rows                               5    5   PASS
    three rows, UPDATE one                           3    3   PASS
    DELETE one of three                              2    2   PASS
    grows 1 -> 2 rows                                2    2   PASS
    NULL-bearing rows, DELETE the (NULL,NULL) row    2    2   PASS
    value round-trip A -> B -> A                     1    1   PASS
    single UPDATE touching 5 rows                    5    5   PASS
    control: UNIQUE key table                        5    5   PASS
    control: PK-column UPDATE                        2    2   PASS
    three byte-identical rows                        3    1   known limit

Parity was re-verified after `OPTIMIZE TABLE ... FINAL` on every table, so
it is a settled state and not a pre-merge artifact. Zero connector errors.

One case cannot reach parity: a keyless table holding two byte-identical
rows. Identical rows are indistinguishable under any sorting key, so
ClickHouse retains one. CollapsingMergeTree and SummingMergeTree track the
multiplicity correctly but collapse the physical rows at merge, so an
ordinary consumer query still sees one row where MySQL has two. The
connector now logs a warning naming the table at create time.

Alternatives, each refuted by measurement rather than argument:

  ORDER BY _version      every record gets a distinct version, so nothing
                         ever deduplicates: an UPDATE leaves both rows and
                         a DELETE tombstone merely adds a third.
  Engine=MergeTree       SELECT ... FINAL then fails with Code: 181
                         ILLEGAL_FINAL, breaking every consumer query and
                         checksum job -- and `SETTINGS final=1` does not
                         fail, it silently returns undeduplicated rows.
  Collapsing/Summing     multiplicity arithmetic is right, but the physical
                         rows collapse to one at merge time.

Two of my own earlier claims are corrected here rather than quietly
dropped. The javadoc added with the UNIQUE-key fix asserted that ordering
by all data columns is "NOT a valid alternative" because an UPDATE would
duplicate the row; the duplication is real, but it follows from emitting
only the after-image, not from the sorting key, and the tombstone removes
it. `CreateTableUniqueKeySortKeyTest.testNoKeyAtAllStillFallsBackToTuple`
asserted the collapse as correct behaviour and is replaced by a test of the
new contract. A UNIQUE key is still preferred over all-columns: it is a
narrower and more meaningful identity.

Twenty-two expectations in MySqlDDLParserListenerImplTest asserted
`ORDER BY tuple()` for keyless tables -- the defective contract itself --
and were rewritten from each test's own column list.

Tests: sink-connector 194 run / 0 failures / 20 errors against an untouched
baseline of 186 / 0 / 20 on the same base commit; the 20 are pre-existing
testcontainers "Could not find a valid Docker environment" errors and the
delta is exactly the 8 new tests. sink-connector-lightweight DDL tests
120 / 1, the one failure being testAutoCreateTable, which also fails on the
untouched baseline. Failing-first verified: disabling only the keyless
fallback fails 5 of the 7 new DDL tests, each citing ORDER BY tuple(),
while the PRIMARY-KEY-wins and UNIQUE-key-wins guards keep passing.


minguyen9988:
Note on the file count: this PR touches **12** files. GitHub shows 13 because the branch is cut from d3f0ba98 (PR #1304's merge commit) while the base here is the 2.10.0 tip, so `sink-connector/python/db_dump/mysql_dumper.py` appears as a pre-existing delta between those two points, not as part of this change. The judge-panel mirror, pinned to the exact base commit, shows the true 12-file diff.

PR #1397, opened 2026-08-24
fix: a REAL column claims Float64 precision it never receives

minguyen9988:
-

### DIFF
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/parser/DataTypeConverter.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/parser/DataTypeConverter.java
@@ -148,18 +148,48 @@ public class DataTypeConverter {
             return bitClickHouseType(columnDefChild);
         }
 
-        // MySQL DOUBLE / DOUBLE PRECISION / FLOAT8 / REAL are 8-byte values
-        // carrying ~15 significant digits. Debezium widens MySQL FLOAT to a
-        // FLOAT64 Kafka schema as well, so FLOAT and DOUBLE are
-        // indistinguishable by schema type alone and the shared type map can
-        // only answer one of them correctly — it answers Float32, silently
-        // truncating every DOUBLE to ~7 significant digits and overflowing
-        // large magnitudes to inf.
+        // MySQL DOUBLE / DOUBLE PRECISION / FLOAT8 are 8-byte values carrying
+        // ~15 significant digits. Debezium widens MySQL FLOAT to a FLOAT64
+        // Kafka schema as well, so FLOAT and DOUBLE are indistinguishable by
+        // schema type alone and the shared type map can only answer one of them
+        // correctly — it answers Float32, silently truncating every DOUBLE to
+        // ~7 significant digits and overflowing large magnitudes to inf.
         // On this DDL path the resolved source JDBC type IS available, so the
         // two can be told apart exactly: FLOAT/FLOAT4 keep Float32, and only
         // the genuine 8-byte types widen to Float64.
-        if (dataType.jdbcType() == Types.DOUBLE || dataType.jdbcType() == Types.REAL) {
+        //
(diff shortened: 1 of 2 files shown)


minguyen9988:
## Summary

A MySQL `REAL` column was created in ClickHouse as `Float64`, but the value arrives at the writer **already narrowed to IEEE-754 single precision**. The column advertised ~15 significant digits while carrying ~7.

Found while auditing whether #1353 is still needed, against the published 2.10.0 image `1304-364c719120e69c99a389460b46c02e2f43b0f00f-lt`.

## Measured — before

MySQL 8.0.36 ROW/FULL/GTID -> ClickHouse 24.8.14.10547. All of `d`, `dp`, `r` are `double` in MySQL's own `SHOW CREATE TABLE`, all three declared `Nullable(Float64)`:

```
column                 mysql                  clickhouse
d  DOUBLE              1234567.8901234567     1234567.8901234567   ok
dp DOUBLE PRECISION    9876543.210987654      9876543.210987654    ok
r  REAL                9876543.210987654      9876543             WRONG
r  REAL                0.1234567890123456     0.12345679          WRONG
```

Row counts are unaffected, so count-based checksum jobs report the table clean. The loss is per-row and silent.

## Root cause is upstream of this connector

Debezium resolves MySQL `REAL` to `Types.REAL` and maps that to a **float32** Kafka schema, while `FLOAT` (`Types.FLOAT`) and `DOUBLE` (`Types.DOUBLE`) both map to float64 — `io.debezium.jdbc.JdbcValueConverters`, debezium 3.1.3.Final, switch keys `6 -> float64`, `7 -> float32`, `8 -> float64`. The narrowing happens before any connector code runs.

**Verified, not assumed.** An earlier revision of this branch made the value bind destination-aware, so a value landing in a `Float64` column bound at double width regardless of its Java class. It was confirmed present in the built jar (`Double.parseDouble` in the compiled `ClickHouseDataTypeMapper`) and made **no difference end to end** — the stored value was still `9876543`. The bits are gone before the bind, so no bind-side change can recover them. That revision was reverted.

## The fix

What the connector *does* control is the declared column type. `REAL` is now declared `Float32`:

```
ClickHouse r == toFloat32(9876543.210987654)  ->  1
```

The stored value is exactly what Debezium delivered. `Float64` only made truncated data look full-precision to every consumer and doubled the stored width for nothing.

Recovering the digits requires a Debezium change (map `Types.REAL` to float64 as `FLOAT` and `DOUBLE` already are); until then `Float32` faithfully describes what the pipeline carries. Happy to take the opposite direction if you would rather carry a connector-side workaround.

`DOUBLE`, `DOUBLE PRECISION` and `FLOAT8` are untouched and still widen to `Float64`; `FLOAT` and `FLOAT4` still resolve to `Float32`.

## Measured — after

```
column   declared            mysql                 clickhouse            result
r        Nullable(Float32)   9876543.210987654     9876543 (=float32)    PASS
d        Nullable(Float64)   9876543.210987654     9876543.210987654     PASS
dp       Nullable(Float64)   9876543.210987654     9876543.210987654     PASS
f8       Nullable(Float64)   9876543.210987654     9876543.210987654     PASS
f        Nullable(Float32)   1.25                  1.25                  PASS
```

Row parity 2/2 after an UPDATE and a DELETE; exactly-representable values (3.25, 1.25) bit-exact on both sides; `Code: 44` = 0, `CANNOT_PARSE` = 0.

## Tests

The `DataTypeConverterTest` case for `REAL` previously asserted `Float64` — it pinned the mismatch — and is corrected here. **Failing-first verified:** with the main-source change stashed, that case fails with `expected:<Float[32]> but was:<Float[64]>`, and passes with the fix.

## Unrelated defect observed, NOT addressed here

A MySQL `FLOAT4` column is dropped from the INSERT entirely and lands NULL in ClickHouse (`Column index missing for column ****f4`). **Reproduced identically on the unmodified published image**, so it is pre-existing and independent of this change. It needs its own fix — happy to open one.

PR #1245, opened 2026-03-01
PostgreSQL → ClickHouse CDC Support — Full Production Pipeline with Verification Toolkit

minguyen9988:
# PR: PostgreSQL → ClickHouse CDC Support — Full Production Pipeline with Verification Toolkit

## Table of Contents

- [Summary](#summary)
- [Motivation / Problem Statement](#motivation--problem-statement)
- [What Changed](#what-changed)
  - [Java Sink Connector](#java-sink-connector)
  - [ANTLR4 Grammars](#antlr4-grammars)
  - [Python Tools (`ch-sink-tools`)](#python-tools-ch-sink-tools)
  - [Integration Tests](#integration-tests)
  - [Deployment Configuration](#deployment-configuration)
- [Architecture](#architecture)
  - [Two-Phase Snapshot + CDC Pipeline](#two-phase-snapshot--cdc-pipeline)
  - [Factory Pattern for MySQL/PostgreSQL Isolation](#factory-pattern-for-mysqlpostgresql-isolation)
  - [PostgreSQL DDL Capture via Event Triggers](#postgresql-ddl-capture-via-event-triggers)
  - [Schema Drift Detection and Auto-Reconciliation](#schema-drift-detection-and-auto-reconciliation)
  - [Checksum Verification Architecture](#checksum-verification-architecture)
  - [Auto-Diff Binary Search](#auto-diff-binary-search)
- [Data Type Coverage](#data-type-coverage)
- [Verification Results](#verification-results)
  - [Checksum Run Progression (Runs 11–28)](#checksum-run-progression-runs-1128)
  - [Final Metrics](#final-metrics)
- [Root Cause Analysis](#root-cause-analysis)
- [Running Instructions](#running-instructions)
  - [1. Build the Java Connector JAR](#1-build-the-java-connector-jar)
  - [2. Build the Python Wheel](#2-build-the-python-wheel)
  - [3. Run the Initial Snapshot (postgres_dumper)](#3-run-the-initial-snapshot-postgres_dumper)
  - [4. Start the CDC Connector](#4-start-the-cdc-connector)
  - [5. Run Checksum Verification](#5-run-checksum-verification)
  - [6. Run Auto-Diff](#6-run-auto-diff)
- [Configuration Reference](#configuration-reference)
- [Breaking Changes](#breaking-changes)
- [Known Issues / Limitations](#known-issues--limitations)
- [Files Changed](#files-changed)
- [Planning Documents](#planning-documents)

---

## Summary

[...]

PR #1398, opened 2026-08-24
PostgreSQL to ClickHouse CDC Support (rebase of #1245 onto 2.10.0)

minguyen9988:
Rebase of #1245 onto `2.10.0`.

This is #1245's work replayed onto the `2.10.0` branch instead of `develop`.
It is opened as a separate branch because updating #1245 in place would
require a force-push over its existing head. Once this is reviewed, #1245
can be closed in favour of this one.

## What changed

| | |
|---|---|
| Base | `develop` → `2.10.0` |
| Commits replayed | 66 (the merge commit from the old base is dropped) |
| Net diff vs `2.10.0` | 100 files, +16,736 / −617 |

`2.10.0` was 120 commits ahead of `develop` at the time of the rebase, so a
number of the PR's changes had to be reconciled against work that landed
independently on the release branch.

## Conflict resolutions worth reviewing

The mechanical resolutions (generated artifacts, `.gitignore` unions) are not
listed. These are the ones that involved a judgment call:

**Base version kept — the release branch already solves the same problem**

- `DebeziumJdbcStorageOperations` view creation. Both branches fixed the
  `renameat2()` `UNSUPPORTED_METHOD` failure on `CREATE OR REPLACE VIEW`.
  `2.10.0` checks `system.tables` and skips creation; #1245 rewrote the DDL to
  `DROP` + `CREATE`. Kept the base's approach.
- `DBMetadata` column metadata. Both switched to querying `system.columns`
  directly; the base's version carries the same fix plus the rationale for why
  `getColumns()` was wrong (JDBC LIKE patterns matching sibling tables).
- `DatabaseOverrideRRMTIT`. The base converted it to polling with
  `try/finally` cleanup; replaying the PR version would have reverted that.
- Two ClickHouse 22.3 tests deleted by the PR stay deleted — the base is on
  Testcontainers 1.21.4, so the deletion rationale still applies.

**Both sides merged**

- DDL cache invalidation in `DebeziumChangeEventCapture`. Kept the base's
  override-map-aware, multi-table invalidation and added the PR's
  schema-drift detector invalidation and its topic-derived table fallback for
  schema-prefixed PostgreSQL topics.
- `replica.status.view`. Kept the base's `%s` parameterization of the
  database/table and the PR's `ts_sec`/`ts_usec` fallback, which is what makes
  the view work against PostgreSQL offsets. Taking either side alone would
  have dropped one of the two.
- `MySqlDDLParserListenerImpl` `TRUNCATE`. Kept the base's qualified-name
  split and applied the PR's backtick quoting of the database name.
- `ClickHouseSinkConnectorConfigVariables`, `ClickHouseDataTypeMapper`,
  `MySqlDDLParserListenerImpl` fields — both sides appended distinct members;
  all were kept.
- `MySqlDDLParserListenerImplTest`: 21 expectation conflicts. The base had
  rewritten the expected DDL (`ORDER BY` / `SETTINGS` / nullability) and added
  new cases, while the PR only added backticks. Kept the base expectations and
  re-applied the backticking (33 lines).
- `BinLogHistoryTest`: the producer now emits both the backticked database
  name and the base's `DateTime64(0)` + `db_time` column, so the expectation
  had to carry both rather than either side as-written.

**One correctness fix beyond a plain replay**

The PR's final commit, "Move sink-connector/python to separate branch",
restores `sink-connector/python` to match 2.9.1 on the assumption the base
carries no such tree. `2.10.0` has since merged the `ch_sink_tools` package
(commits `3a79f1b2`, `e142d5f7`). Replayed verbatim, that commit deleted 58
files belonging to the base — the python package would have been removed from
the release branch as a side effect of this PR.

The commit was amended so the subtree is restored to the **new** base rather
than the old one. `sink-connector/python` is now byte-identical to `2.10.0`,
which preserves the commit's actual intent (keep python out of this PR's
scope) without deleting anything. Verified: `git diff origin/2.10.0 HEAD --
sink-connector/python` is empty.

## Verification

- No conflict markers remain in any tracked file.
- The branch's file footprint against `2.10.0` contains no files the original
  PR did not touch; 9 of the original files dropped out, each because the base
  supersedes them (listed above).
- Hand-merged Java files pass a brace/paren balance check that ignores
  literals and comments.

**Not verified:** the branch has not been compiled or tested. No JDK 17 was
available on the machine used for the rebase (JDK 8 only), so `mvn verify`
could not run. CI on this PR is the first real compile — the semantic merges
above are the places to look if it fails.


minguyen9988:
## Conflict resolved — now `MERGEABLE`

Head `60e3bafb`, base `2.11.0`. Three files conflicted. Two of the three needed content in **neither** parent, because taking either side verbatim would have dropped a fix the other side depends on.

### `MySqlDDLParserListenerImpl` — the keyless row identity

This branch carried the `_row_key` MATERIALIZED fingerprint from #1399 (`ORDER BY cityHash64(...)` over every column, `PRIMARY KEY tuple()`). The base has since **reverted** that approach in #1401: a column named by the sort-key expression cannot be dropped (Code: 44), and a column added later is absent from the expression, so two rows differing only in the new column collapse to one. #1401 replaced it with MySQL's generated invisible primary key plus `KeylessTablePreflight`.

Taking "ours" here would not merely have reinstated a known-broken identity — **it would not have compiled**: #1401 removed the `rowKeyColumnDefinition` local that this branch's block reads. Dropped, base kept.

The branch's *other* contribution in the same hunk — the ALIAS column emission from `column_type_override.alias.*` — exists only here and is preserved.

### `MySqlDDLParserListenerImpl` — `postProcessModifyColumn`

Both sides fixed something real and neither subsumes the other:

| side | fix |
|---|---|
| base | `RENAME COLUMN` to `RENAME COLUMN IF EXISTS` (#1402), so replaying an already-applied rename is a no-op instead of a Code: 10 that stalls the whole stream — DDL is retried indefinitely |
| branch | quotes the database identifier, so a database whose name needs quoting parses |

Taking base loses the quoting; taking ours loses the replay guard. Merged as both.

### `DebeziumChangeEventCapture`

Additive on this branch only (`PostgresConnectorConfig` initialisation); the base has no competing line. Branch kept.

### `MySqlDDLParserListenerImplTest` — 59 hunks

Every hunk is the same shape: the base expectation carries semantic guards this branch predates (`CREATE TABLE IF NOT EXISTS`, `RENAME`/`DROP COLUMN IF EXISTS`, the #1401 sort-key revert), and this branch's carries the quoted database name. Resolved uniformly — base expectation, quoting re-applied — never by picking a side. 46 of the 59 were equivalent once both transformations are accounted for.

### Eleven stale expectations fixed while here

The branch's quoting change never reached the assertions that sit **outside** the conflict hunks, so they still spelled the database bare and failed against this branch's own generator — on the pre-merge head, not only after the merge. Verified on an untouched worktree at `4fb183f5`, where `MySqlDDLParserListenerImplTest` is **96/107**.

Fixed: `testAlterAddColumnWithInterleavedAlgorithmClauses`, `testAlterAddColumnWithInterleavedLockClause`, `testAlterAddColumnWithTrailingHintsLeavesNoDanglingComma`, `testAlterDatabaseModifyColumns`, `testChangeColumnWithExplicitNull`, `testChangeColumnWithNotNull`, `testModifyColumnWithExplicitNull`, `testModifyColumnWithNotNull`, `testPartitionByRange`, `testCreateTableWithPartitionByRange`, `truncateTableWithQualifiedName`, and `CreateTableIdempotentTest.testGuardPlacement` (a base test whose table-name lookup this branch's quoting broke).

## Verification

Build: both modules, `mvn clean install -DskipTests`, green.

**Surefire cannot discover tests in this module on the build host, and it fails open.** `junit-platform-engine` 1.8.2 lands on the test classpath while jupiter needs 1.9.1, so discovery dies with `NoSuchMethodError: getDiscoveryListener()` and surefire then reports `Tests run: 0` as a **PASS**. Reproduced byte-identically on the untouched baseline, so it is the host and not this merge — but it also means a green surefire run here would prove nothing. Executed instead through the JUnit Platform launcher directly, with the stale 1.8.2 jars removed from the classpath:

```
                                  baseline (4fb183f5)   this merge
MySqlDDLParserListenerImplTest         96 / 107          107 / 107
all unit classes                      264 / 278          314 / 317
```

The 3 remaining failures are pre-existing on this branch and untouched by the merge:

- three `PostgresSchemaReconcilerTest` timestamp cases expect a bare `Nullable(DateTime64(N))` while this branch's own commits `a7049bde` / `78f2618d` made `ClickHouseDataTypeMapper` emit `Nullable(DateTime64(N, 'UTC'))` — both commits are on this branch and on neither base, so this is a genuine internal inconsistency worth a look, not merge fallout;
- one testcontainers case with no Docker on the host.

Left for you rather than papered over inside a merge commit.

Docker-dependent integration suites were not run locally; CI exercises them.

minguyen9988:
## Postgres end-to-end run — real containers, and it found a second pre-existing defect

Head is now `fa5654bb`. I stood up the actual stack rather than trusting the unit suite: `debezium/postgres:15-alpine` → sink connector → `clickhouse-server:24.8`, driven by the repo's own testcontainers ITs against a rootless podman socket. Runs reach `Snapshot completed` and auto-create the target tables, so every assertion below is measured against DDL the connector actually produced.

### What it found

`PostgresInitialDockerIT` failed at the `created` column. The connector emits:

```
CREATE TABLE `public`.`tm`(...
    `set_date` Nullable(DateTime64(6, 'UTC')),
    `created`  Nullable(DateTime64(6, 'UTC')),
    `updated`  Nullable(DateTime64(6, 'UTC')), ...)
```

while the test asserted `Nullable(DateTime64(6))`. The test has never matched this branch.

**Not merge fallout — proven by A/B, not asserted.** The same IT fails identically at the same line on the untouched pre-merge head `4fb183f5`, in a separate worktree against the same podman socket. The zone tagging comes from commit `78f2618d`, which is on this branch and in *neither* base.

**The production code is right and I did not touch it.** Debezium timestamps are UTC by definition, and `78f2618d` tagged the zone in `ClickHouseDataTypeMapper.mapDebeziumSchemaToDDL` — the record-schema auto-create path, which is the path PostgreSQL uses. Dropping the zone would make an instant read as local wall-clock time to every consumer. Only the expectations were left behind.

Fixed five assertions, one per Postgres IT, all on the same `created` column. Scope is deliberately the record-schema path only: the MySQL DDL path resolves through `DataTypeConverter`, not this mapper, so `CreateTableDataTypesIT` and friends still legitimately expect a bare `DateTime64` and are untouched. **No production file changes in this commit.**

### Full suite result

| | |
|---|---|
| `PostgresInitialDockerIT` | 0/1 → **1/1** |
| Full 8-class Postgres IT suite | 31 started, **30 pass**, 1 fail |

The one failure was `PostgresSchemaDriftIT.testMultipleColumnsAdded` ("Row 200 must be present in ClickHouse"). Re-run **alone** on this same merged tree it passes **2/2**, and it also passes 2/2 alone on the baseline — so that is shared-JVM/container contention across a long batch, not a defect in the diff. Worth knowing if it shows up in CI, which runs these back-to-back the same way.

Suites exercised: `ClickHouseDebeziumEmbeddedPostgresPgoutputDockerIT`, `...DecoderBufsDockerIT`, `PostgresDDLOperationsIT`, `PostgresDeleteOperationsIT`, `PostgresUpdateOperationsIT`, `PostgresSchemaDriftIT`, `PostgresSchemaAwareNamingIT`, `PostgresPgoutputMultipleSchemaIT`.

### Still open for you

The unit-suite counterpart of the same question: three `PostgresSchemaReconcilerTest` timestamp cases expect the bare `Nullable(DateTime64(N))` while `ClickHouseDataTypeMapper` emits the UTC-tagged form. I left those deliberately — it is the same "which side is authoritative" call, and the answer there may reasonably be to align the reconciler rather than the test. Your branch, your call.

minguyen9988:
**2. The null-record early return — this one is load-bearing for #1379.**

This branch had:

```java
chStruct = ...parse(...);
if (chStruct == null) {
    log.debug("Skipping record with null ClickHouseStruct ...");
    return null;                  // returns out of the method
}
chStruct.setSequenceNumber(sequenceNumber);
```

That predates both #1416 and #1428 and is **incompatible** with the fix the base now carries. A heartbeat parses to null by contract; under #1428 the caller (`handleChangeEventBatch`) must still *see* that record so `commitControlRecordOffset` can acknowledge its offset — on an idle source the post-snapshot state rides exclusively on heartbeats. Returning early means the record never reaches the control-record path, `replica_source_info` stays pinned at `snapshot=INITIAL / snapshot_completed=false` forever, and every restart re-runs the whole snapshot. That is precisely the reported bug, so the early return was dropped in favour of the base and a comment now marks the spot.

Worth flagging: **#1416 and #1428 were not reachable from this PR's previous head.** They are on `develop`, and `2.11.0` carries them as cherry-picks rather than ancestors. Before this merge the branch still had the pre-#1416 code at the parse site and none of the #1428 machinery (`handleChangeEventBatch` / `commitControlRecordOffset` / `isPipelineQuiescent`). This merge is what brings the #1379 fix onto the branch.

### End-to-end verification of #1379

Real containers, not the unit suite: `postgres:17.5` (`wal_level=logical`) → connector → `altinity/clickhouse-server:24.8.14.10547`, seeded to the reporter's shape (`AdditionalConsumptions` 35,044 rows, `AdditionalConsumers` 1, `__EFMigrationsHistory` 16, three empty tables, publication over all). Same stack, same config, **only the JAR swapped**.

JAR provenance verified in bytecode before running, so the comparison cannot be mislabelled:

```
premerge.jar  javap -> 0/3 of handleChangeEventBatch, commitControlRecordOffset, isPipelineQuiescent
merged.jar    javap -> 3/3
```

| | pre-merge (control) | merged |
|---|---|---|
| snapshot batches, run 1 | 20 | 20 |
| rows landed in ClickHouse | 35,044 | 35,044 |
| **batches on RESTART (idle DB)** | **20** | **0** |
| **records re-sent on restart** | **35,061** | **0** |

Offset written after the snapshot — this is the whole bug:

```
pre-merge  {"last_snapshot_record":false,"lsn":31676656,"txId":750,...,"snapshot":"INITIAL","snapshot_completed":false}
merged     {"lsn":31991872,"txId":762,"ts_usec":1788420081942147}
```

The control reproduces @tomh4's report exactly, and after its restart it is *still* `snapshot_completed=false` — it would loop indefinitely on an idle source. The merged build writes a clean, snapshot-free offset; its restart replayed **nothing** and sat in streaming acknowledging heartbeats at `lsn=31991872`, which are exactly the records #1428 commits.

Because the control fails and the merged build passes on the same harness, this is a failing-first result rather than a harness that silently did not run.

Also confirmed the second half of #1428 on the live table: three physical offset rows with unique random `id`s collapse via `ORDER BY offset_key` + `FINAL` to exactly one row, and all 27 shipped configs use `ORDER BY offset_key` (the one remaining `ORDER BY id` is an unrelated test-fixture table in `sink-connector/tests/integration/tests/steps/sql.py`, not the offset store).

### Build

`mvn clean install -DskipTests`, both modules, green under JDK 17.0.2 / Maven 3.9.11.

### Not covered

The full Postgres IT suite has **not** been re-run against this merged head. The last run, on `fa5654bb`, was 30/31 — the single failure `PostgresSchemaDriftIT.testMultipleColumnsAdded` passes in isolation and was attributed to container contention. CI exercises these.

Still open for the branch owner, unchanged from before: three `PostgresSchemaReconcilerTest` timestamp cases expect the bare `Nullable(DateTime64(N))` while `ClickHouseDataTypeMapper` emits the UTC-tagged form.

PR #1399, opened 2026-08-24
fix: a keyless table's all-column sorting key freezes its schema

minguyen9988:
-


minguyen9988:
## Summary

#1394 gave a MySQL table with no PRIMARY KEY and no UNIQUE key an all-column ClickHouse sorting key, so ReplacingMergeTree had a row identity to deduplicate on. That fixed the collapse, but ClickHouse forbids altering any column that participates in the sorting key, so it left every data column unalterable and the table schema frozen.

## What breaks

Measured on ClickHouse 24.8.14.10547 against a table keyed that way:

    ALTER TABLE t MODIFY COLUMN b Nullable(Int32)   Code: 524 ALTER_OF_COLUMN_IS_FORBIDDEN
    ALTER TABLE t RENAME COLUMN b TO b_new          Code: 524 part of key expression
    ALTER TABLE t DROP COLUMN v                     Code: 47  UNKNOWN_IDENTIFIER
    ALTER TABLE t ADD COLUMN c ...                  OK
    MODIFY on a column NOT in the key (control)     OK

The connector retries a failed DDL ten times then gives up, so the change never lands and replication of that table schema is stuck. Code 524 over one AlterTableChangeColumnIT run: 0 before #1394, 58 after.

Bisected: that IT passes on d8b00eb3 (pre-#1394) and errors on both 33ff7077 and 78b868c3, so #1394 is the culprit, not #1395 or #1396.

A fix for silent data loss must not cost the table its ability to take any other operation.

## The fix

Row identity moves into one generated column instead of the data columns:

    _row_key UInt64 MATERIALIZED cityHash64(concat(
        toString(isNull(a)), ifNull(toString(a),''), 0x01 separator,
        toString(isNull(b)), ifNull(toString(b),'')))
    PRIMARY KEY tuple() ORDER BY _row_key

MATERIALIZED, so ClickHouse computes it and it never enters the writer INSERT column list.

Each column contributes an explicit null-flag as well as its text, so a genuine NULL cannot collide with any literal the column might hold (NULL, empty string and '0' all hash differently). The flag also keeps the expression non-Nullable, which a sorting key requires: hashing the raw Nullable columns yields Nullable(UInt64) and the insert dies with Code 349, and cityHash64(tuple(...)) throws Code 48 on any NULL member. Both measured before settling on this form. The 0x01 separator keeps (ab,c) distinct from (a,bc).

An increasing counter was considered and rejected: it cannot recognise a re-sent row, so an UPDATE would duplicate and a DELETE would not match. The fingerprint is the only shape that preserves deduplication.

allow_nullable_key is no longer emitted on any path: the fingerprint is non-Nullable by construction, a PRIMARY KEY is NOT NULL by MySQL own rule, and a UNIQUE key is adopted only when every column is NOT NULL.

## Second-order defect fixed with it

PreparedStatementExecutor decides whether to tombstone the before-image by comparing sorting-key columns. No CDC record carries _row_key, so that comparison would skip it and report not-relocated for every UPDATE, stranding the pre-update row and reintroducing exactly the loss the fingerprint exists to prevent. For that key any change to any value moves the row by construction, so the full before/after images are compared instead.

## Measured after

    4 distinct rows incl (NULL,NULL) and (NULL,q)   -> 4    correct
    re-send at a higher version (UPDATE)            -> 4    no duplicate
    tombstone the (NULL,NULL) row                   -> 3    correct
    MODIFY / RENAME / ADD / DROP COLUMN             -> all succeed
    rows intact after that schema churn             -> 3
    3 byte-identical rows                           -> 1    unchanged limit

The byte-identical-rows limit is unchanged: identical rows are indistinguishable under any sorting key, fingerprint or not.

AlterTableChangeColumnIT and AlterTableModifyColumnIT both pass, run against real containers, Code 524 count 0.

## Also corrected: a stale assertion

MySqlDDLParserListenerImplTest#testAutoCreateTable expected Doublex Nullable(Float32). #1386 correctly made MySQL DOUBLE map to Float64, but this expectation was never updated, so it has been asserting the pre-#1386 defect and failing ever since, including on the unmodified parent. Field-by-field diff showed exactly one mismatch out of 42. assertTrue changed to assertEquals so the next failure prints a readable diff.

## Tests

MySqlDDLParserListenerImplTest 108 run / 0 failures (was 1 on the unmodified parent). CreateTableNoKeySortKeyTest 7/0, CreateTableUniqueKeySortKeyTest 9/0, DataTypeConverterTest 1/0.

## Note on existing keyless tables

This changes the shape of newly auto-created keyless tables. Tables already created with the all-column key keep it and stay schema-frozen until recreated: worth calling out in the release notes.

minguyen9988:
## Update: two blockers fixed, two still OPEN

An adversarial review of the first commit found four defects. I reproduced all four live on 24.8.14.10547 rather than taking or dismissing them on argument. All four were real. Two are now fixed in 7abf4be3; **two remain open and this PR should not be merged until they are.**

### Fixed in 7abf4be3

**1. The delimiter-joined fingerprint was not an unambiguous encoding.** Values are not escaped, so a value containing the 0x01 separator shifts the boundary and spoofs the next column null-flag:

    cityHash64(concat('0','x','\x01','0','\x010z'))
      = cityHash64(concat('0','x\x010','\x01','0','z'))   -> 1  COLLISION

The rows ('x', 0x01+'0z') and ('x'+0x01+'0', 'z') are distinct in MySQL but hash identically, so RMT collapses one. Silent data loss, the exact failure the fingerprint exists to prevent. Now passed as a tuple, whose members keep their boundaries: same payload pair returns 0. NULL-safety re-verified unchanged (NULL, empty and '0' still hash to three different values).

**2. The generated name could collide with a source column.** A keyless table declaring its own _row_key rendered two definitions of that name; ClickHouse rejects it with Code 44 and replication of that table stalls. Now disambiguated by prepending underscores, the same way the delete-marker column already does.

### STILL OPEN — do not merge

**3. DROP COLUMN on a column that feeds the fingerprint is rejected.** My earlier DROP test was flawed: it dropped a column added *after* creation, which is absent from the fingerprint. Dropping one that feeds it:

    Code: 44. Cannot drop column `b`, because column `_row_key` depends on it. (ILLEGAL_COLUMN)

So the schema freeze is *narrowed*, not eliminated: MODIFY and RENAME now work, DROP still does not.

**4. ADD COLUMN does not extend the fingerprint, and that collapses rows.** The expression is fixed at CREATE time. After an ADD COLUMN, two rows differing only in the new column share a fingerprint:

    2 distinct rows (differ only in the new column) -> 1

Measured, not theorised.

Both need the connector to rebuild the fingerprint as part of replicating ADD/DROP COLUMN (ALTER TABLE ... MODIFY COLUMN _row_key ... alongside the column change). I am working that now and will push it here.

Posting this rather than sitting on it, because the branch currently looks green on the unit tests while these two remain.

minguyen9988:
Superseded by #1401 and should be closed. Same reason as #1400: this PR invented a row identity downstream, which blocks DROP COLUMN on the table. #1401 uses MySQL's generated invisible primary key (my_row_id) as the source of truth and refuses to replicate a table for which MySQL cannot supply one.

PR #1400, opened 2026-08-25
fix: a keyless table's generated row key breaks its checksum comparison

minguyen9988:
Fixes the 10 CI failures on the `2.10.0` head (`78b868c3`), all of which trace back to #1394.

## The regression

#1394 gave a keyless source table (no PRIMARY KEY, no UNIQUE key) an all-column ClickHouse sorting key so ReplacingMergeTree would stop collapsing its rows. That **froze the schema of every such table** — ClickHouse forbids altering any column in the sorting key. Bisected across three commits on ClickHouse 24.8.14.10547:

```
MODIFY COLUMN b Nullable(Int32)   Code: 524 ALTER_OF_COLUMN_IS_FORBIDDEN
RENAME COLUMN b TO b_new          Code: 524 "part of key expression"
DROP COLUMN v                     Code: 47  UNKNOWN_IDENTIFIER
MODIFY on a non-key column        OK   (control)

Code: 524 count:  d8b00eb3 = 0    33ff7077 = 14    78b868c3 = 58
```

The connector retries the DDL 10x (~45s) and gives up, so the schema change never lands — in CI **and in production**.

## The fix

The identity is a plain `_row_key UInt64` whose value the **connector** computes per record, not a `MATERIALIZED` expression. An expression has to name the columns, which bakes the column list into the table and is brittle in both directions — both measured live:

```
DROP COLUMN of a named column
  -> Code: 44 "Cannot drop column b, because column _row_key depends on it"
ADD COLUMN, then 2 rows differing ONLY in the new column
  -> both hash the same, ReplacingMergeTree collapses them: 2 rows in, 1 out
```

Computing it connector-side removes the dependency entirely: the fingerprint covers whatever fields the record carries, so ADD / DROP / MODIFY / RENAME COLUMN all succeed with no maintenance of the key.

An auto-increment counter was rejected: it cannot recognise a re-sent row, so UPDATE would duplicate and DELETE would not match.

## Four further defects, each a silent data loss or corruption

Found by testing end to end, not by unit tests. All were reproduced before being fixed.

| # | defect | consequence |
|---|---|---|
| 1 | `_row_key` was never in the INSERT column list | every row written with the UInt64 default `0`, all rows share a sorting key, **whole table collapses** — measured 4 rows in, 1 out |
| 2 | fingerprint unstable across a restart | `String.valueOf` on a `byte[]`/`ByteBuffer` returns the JVM identity hash, so a BLOB row re-sent after a **connector restart or host move** duplicates instead of replacing |
| 3 | a SOURCE column named `_row_key` was overwritten with the fingerprint | user data replaced; row counts unchanged, so **invisible to a count-based check** |
| 4 | the checksum tooling did not exclude `_row_key` | the two sides hash different column sets, so **every keyless table is reported as a MISMATCH** by the Jenkins `*_top_level_table_checksum_*` jobs — the same signal a genuine divergence produces |

Defect 1 before/after:

```
INSERT INTO `keyless_evolves`(`a`,`b`,`_version`,`is_deleted`)                     <- _row_key missing
INSERT INTO `keyless_binary`(`a`,`payload`,`_row_key`,`_version`,`is_deleted`)     <- fixed (from the IT log)
```

Defect 4 spans `db_compare` and the `ch_sink_tools` mirror. The existing "don't exclude X when `_X` is also present" rule then keeps a *source* `_row_key` column in the comparison — the opposite risk, and the worse one, since dropping it would turn a genuine divergence into a false PASS.

Also stops logging the generated column as `"not present in source"` once per column per record, which made a working table look broken (hundreds of lines per run to 0).

## Tests

Every result below is a live measurement against real MySQL + ClickHouse containers.

**`KeylessRowIdentityIT`** (new, 5 tests, all passing in 778s):

```
Scenario                                     Result
3 distinct rows inserted                     3 rows in ClickHouse      PASS
UPDATE (same row, new value)                 replaced, no duplicate    PASS
DELETE tombstone                             row removed               PASS
Connector RESTART, BLOB rows re-sent         replaced, no duplicate    PASS
ADD / MODIFY / RENAME / DROP COLUMN          all succeed               PASS
Rows differing only in a post-ADD column     stay distinct             PASS
Source column named _row_key                 no collision, value kept  PASS
Keyed table (control)                        untouched                 PASS
```

`DDLBaseIT` stops its **static** ClickHouse container in `@AfterEach`. Every other subclass has exactly one test, so that never showed; this class has five and the second would have run against a stopped container. Both containers are recreated per test here, which also gives each test its own Debezium offset storage so the restart test cannot resume another test's offsets.

**`test_keyless_checksum_replication.py`** (new) — runs the **real `db_compare` checksum scripts**, the same ones the Jenkins compare jobs run, across the initial seed, an UPDATE/DELETE, and an ADD/MODIFY COLUMN, asserting md5 **and** row count agree. It also diverges a source `_row_key` column on purpose to prove over-broad exclusion would not hide it. Wired into `sink-connector-lightweight-checksum-tests.yml`.

**`keyless_row_key_exclusion_test.py`** (new, 4 tests) — pins the exclusion rules. **Failing-first verified**: with the exclusion reverted, 3 of 4 fail.

**Unit tests**: `MySqlDDLParserListenerImplTest` 108, `CreateTableNoKeySortKeyTest` 7, `CreateTableUniqueKeySortKeyTest` 9, `DataTypeConverterTest` 1 — all green. Includes the stale `Doublex Float32 to Float64` assertion left by #1386, which also failed on the baseline.

Supersedes #1399.


minguyen9988:
Superseded by #1401 and should be closed. This PR derived the row identity downstream (a connector-computed fingerprint). That cannot work: it blocks DROP COLUMN, because a key built out of the data columns necessarily depends on them. #1401 takes the identity from MySQL instead — the generated invisible primary key (my_row_id, 8.0.30+), which is a real column carried by the binlogged DDL and by every row image, so schema evolution stays free. Where MySQL cannot supply one, #1401 refuses to replicate rather than producing a ClickHouse copy that silently disagrees with its source.

PR #1402, opened 2026-08-25
fix: make generated schema-change DDL safe to replay after a restart

minguyen9988:
Fixes a replication-stalling bug found while verifying the `2.10.0` image end to end, plus a documentation defect that makes correct behaviour look like data corruption. Both are **pre-existing** and unrelated to the recent keyless/GIPK work.

## Bug 1 — a restart after a schema change stalls the whole stream

Debezium flushes offsets periodically, so a connector restart re-delivers every DDL event committed since the last flush — including events **already applied** to ClickHouse. The generated statements carried no existence guard, so the replay failed. Because DDL is retried indefinitely, one failed replay stalls the **entire replication stream**, not just the offending table, and the ClickHouse copy silently falls behind its source.

Observed end to end: a connector restarted after a `DROP COLUMN` retried the replayed statement 20 times and stopped applying row events altogether — MySQL had 3 rows, ClickHouse 2.

Reproduced directly on ClickHouse 24.8.14, which showed the problem is **wider than the `DROP` that surfaced it**:

```
replayed ADD COLUMN     Code: 15 DUPLICATE_COLUMN
                        "column with this name already exists"
replayed DROP COLUMN    Code: 10 NOT_FOUND_COLUMN_IN_BLOCK
                        "Cannot find column `x` to drop"
replayed RENAME COLUMN  Code: 10 NOT_FOUND_COLUMN_IN_BLOCK
                        "Cannot find column `x` to rename"
replayed MODIFY COLUMN  succeeds
```

So `ADD` / `DROP` / `RENAME` are now guarded, and **`MODIFY` deliberately is not**: re-applying the same type is already a no-op, and adding `IF EXISTS` there would swallow a modification of a column that genuinely does not exist — turning a real schema divergence into silence. A test pins that asymmetry so it is not "tidied up" later.

### `RENAME` needed more than the constant

The rename is emitted from **two paths that bypass** `Constants.RENAME_COLUMN`: `postProcessModifyColumn` built the statement from a hardcoded literal, and `parseRenameColumn` echoes the source tokens verbatim (so the guard is injected as the `COLUMN` keyword goes past). Fixing only the constant would have left both unguarded — the grep-and-fix-one-site trap.

### Not a regression

`DROP_COLUMN` has been unguarded since `e56f0418` (2025-04-04); the recent keyless work never touched it. The same file already carries the commit *"Added if exists to drop constraint"* (`7d186e66`) — this completes a fix that was started for one operation and never applied to its siblings.

## Bug 2 — `persist.raw.bytes` documentation was misleading

BLOB/BINARY/BIT values are stored **hex-encoded as text** by default. The MySQL literal `x'0102030405'` becomes the 10-character string `0102030405`, so `hex(col)` in ClickHouse returns `30313032303330343035` — the hex *of that text* — and a byte-for-byte comparison against the source does not match.

That is a configured default, not corruption, but the old one-line comment described only the `true` branch and never said what the default actually does. A keyed control table renders identically, confirming it is independent of row identity. The comment now documents both modes, the exact consequence, and when to flip it (e.g. when `db_compare` checksum jobs compare binary values). **The default is unchanged** — flipping it would require rewriting existing tables.

## Tests — 145 green

`DdlReplayIdempotencyTest` (new, 5) pins a guard on each generated statement, on **every clause of a multi-clause statement** (guarding only the first leaves the same defect with a narrower trigger), and on the deliberate `MODIFY` exception.

**Failing-first verified** — with the guards reverted, 4 of 5 fail, each naming the exact production error code:

```
testAddColumnIsGuarded      ... Code: 15 DUPLICATE_COLUMN
testDropColumnIsGuarded     ... Code: 10, retry loop stalls replication
testRenameColumnIsGuarded   ... Code: 10, old name no longer exists
testEveryClause...Guarded   ... expected:<2> but was:<0>
```

Pinned expectations in `MySqlDDLParserListenerImplTest` were updated for the guarded output. The MySQL **input** statements in those tests are deliberately unchanged — only the ClickHouse expectations moved.

### DIFF
--- a/sink-connector-lightweight/docker/config.yml
+++ b/sink-connector-lightweight/docker/config.yml
@@ -117,7 +117,24 @@ schema.history.internal.jdbc.table.name: "altinity_sink_connector.replicate_sche
 # enable.snapshot.ddl: If set to true, the connector will parse the DDL statements from the initial load
 enable.snapshot.ddl: "true"
 
-# persist.raw.bytes: If set to true, the connector will persist raw bytes as received in a String column.
+# persist.raw.bytes: how BLOB/BINARY/VARBINARY/BIT values are written. Both
+# modes target a String column; they differ in what that column contains.
+#
+#   false (default) -- the bytes are HEX-ENCODED and the hex TEXT is stored.
+#                      MySQL x'0102030405' becomes the 10-character string
+#                      "0102030405", so hex(col) in ClickHouse returns
+#                      30313032303330343035 (the hex OF THAT TEXT), not
+#                      0102030405. A byte-for-byte comparison against the
+#                      source therefore does not match, and the column takes
+#                      twice the space. The value is recoverable with
+#                      unhex(col).
+#   true            -- the raw bytes are stored unchanged, so the column is
+#                      byte-identical to MySQL and hex(col) matches HEX(col)
+#                      on the source. Use this when binary values are compared
+#                      against MySQL, e.g. by the db_compare checksum jobs.
+#
+# The default stays false for backward compatibility: flipping it changes the
+# stored representation, so existing tables would need to be rewritten.
 persist.raw.bytes: "false"
 
 # auto.create.tables: If set to true, the connector will create tables in the target based on the schema received in the incoming message.
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/Constants.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/Constants.java
@@ -72,40 +72,68 @@ public class Constants {
     public static final String NOT_NULLABLE = "NOT NULL";
 
     /**
-     * Template for adding a column, e.g. "ADD COLUMN %s %s".
+     * Template for adding a column, e.g.
+     * "ADD COLUMN IF NOT EXISTS %s %s".
+     *
+     * <p>The existence guard is what makes DDL replay survivable. Debezium
+     * flushes offsets periodically, so a restart re-delivers every DDL event
+     * committed since the last flush -- including ones already applied to
+     * ClickHouse. Without the guard the replayed statement fails, and because
+     * DDL is retried indefinitely that failure STALLS THE ENTIRE REPLICATION
+     * STREAM, not merely the offending table. Measured on 24.8.14: a replayed
+     * unguarded ADD COLUMN returns
+     * {@code Code: 15 DUPLICATE_COLUMN "column with this name already exists"}.</p>
      */
-    public static final String ADD_COLUMN = "ADD COLUMN %s %s";
+    public static final String ADD_COLUMN = "ADD COLUMN IF NOT EXISTS %s %s";
 
     /**
      * Template for adding a nullable column, e.g.
-     * "ADD COLUMN %s Nullable(%s)".
+     * "ADD COLUMN IF NOT EXISTS %s Nullable(%s)".
+     *
+     * <p>Guarded for the same reason as {@link #ADD_COLUMN}.</p>
      */
     public static final String ADD_COLUMN_NULLABLE =
-            "ADD COLUMN %s Nullable(%s)";
+            "ADD COLUMN IF NOT EXISTS %s Nullable(%s)";
 
     /**
      * Template for modifying a column, e.g. "MODIFY COLUMN %s %s".
+     *
+     * <p>Deliberately unguarded: MODIFY COLUMN is already idempotent, since
+     * re-applying the same type is a no-op. Verified on 24.8.14 -- a replayed
+     * MODIFY COLUMN succeeds. Adding IF EXISTS here would additionally SWALLOW
+     * a modification of a column that genuinely does not exist, turning a real
+     * schema divergence into silence.</p>
      */
     public static final String MODIFY_COLUMN = "MODIFY COLUMN %s %s";
 
     /**
      * Template for modifying a column to be nullable, e.g.
      * "MODIFY COLUMN %s Nullable(%s)".
+     *
+     * <p>Unguarded for the same reason as {@link #MODIFY_COLUMN}.</p>
      */
     public static final String MODIFY_COLUMN_NULLABLE =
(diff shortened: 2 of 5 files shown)


minguyen9988:
## How each bug is fixed

Posting a per-bug walkthrough of the change, since the two fixes are very different in kind: the first changes behaviour, the second only changes a comment.

---

## Bug 1 — a restart after a schema change stalls the whole stream

### What actually goes wrong

Debezium flushes offsets *periodically*, not per event. So when the connector restarts, every DDL event committed since the last flush is delivered again — including events already applied to ClickHouse. Replaying an already-applied schema change fails, and because DDL is retried indefinitely, that one failure never drains: it blocks the queue behind it and **the entire replication stream stops**, not just the offending table.

That last part is what makes it severe. Row events for every other table stop being applied too, and nothing errors loudly at the pipeline level — ClickHouse simply falls further behind its source.

Observed end to end during the `2.10.0` image verification: a connector restarted after a `DROP COLUMN` retried the replayed statement 20 times and stopped applying row events; MySQL had 3 rows, ClickHouse 2.

### The scope was wider than the reported `DROP`

The report named `DROP COLUMN`. Before fixing it I replayed each schema operation directly against ClickHouse 24.8.14 to find out which others share the defect:

```
replayed ADD COLUMN     Code: 15 DUPLICATE_COLUMN
                        "column with this name already exists"
replayed DROP COLUMN    Code: 10 NOT_FOUND_COLUMN_IN_BLOCK
                        "Cannot find column `x` to drop"
replayed RENAME COLUMN  Code: 10 NOT_FOUND_COLUMN_IN_BLOCK
                        "Cannot find column `x` to rename"
replayed MODIFY COLUMN  succeeds
```

`ADD` and `RENAME` stall identically. Fixing only `DROP` would have left two live instances of the same bug, each waiting for a restart that happened to follow that operation instead.

### The fix

Existence guards on the three statements that are not self-idempotent:

| operation | before | after |
|---|---|---|
| `ADD COLUMN` | `ADD COLUMN %s %s` | `ADD COLUMN IF NOT EXISTS %s %s` |
| `DROP COLUMN` | `DROP COLUMN %s` | `DROP COLUMN IF EXISTS %s` |
| `RENAME COLUMN` | `RENAME COLUMN` | `RENAME COLUMN IF EXISTS` |
| `MODIFY COLUMN` | `MODIFY COLUMN %s %s` | **unchanged — deliberately** |

The guard costs nothing in the normal path: dropping a column that is already gone reaches exactly the intended end state. The difference is that the replay becomes a no-op instead of a permanent stall.

### Why `MODIFY COLUMN` is deliberately left alone

Re-applying the same type is already a no-op, so it needs no guard — verified above. Adding `IF EXISTS` there would do something actively harmful: it would also swallow a modification of a column that genuinely does not exist, converting a real schema divergence into silence. That is the opposite failure of the one being fixed here, and the worse of the two, because a stall is at least visible.

`testModifyColumnIsDeliberatelyNotGuarded` pins that asymmetry so a later cleanup pass doesn't "make it consistent".

### `RENAME` needed three fixes, not one

Patching the constant is the obvious move and it is not sufficient — the rename is emitted from two other paths that never read it:

1. **`Constants.RENAME_COLUMN`** — the constant.
2. **`postProcessModifyColumn`** — built the statement from a hardcoded string literal, bypassing the constant entirely.
3. **`parseRenameColumn`** — echoes the source parse-tree tokens verbatim (`RENAME`, `COLUMN`, the two names), so there is no template to change; the guard is injected as the `COLUMN` keyword goes past.

Fixing only (1) would have left (2) and (3) unguarded, and the unit tests would still have gone green. This is why the change touches the parser and not just the constants file.

### Not a regression

`DROP_COLUMN` has been unguarded since `e56f0418` (2025-04-04). The recent keyless/GIPK work never touched it — I checked the file history rather than assuming. The same file already carries the commit *"Added if exists to drop constraint"* (`7d186e66`), so this completes a fix that was started for one operation and never applied to its siblings.

---

## Bug 2 — `persist.raw.bytes`: a documentation fix, not a code fix

### What is actually happening

BLOB/BINARY/BIT values are stored **hex-encoded as text**. The MySQL literal `x'0102030405'` becomes the 10-character *string* `0102030405`, so `hex(col)` in ClickHouse returns `30313032303330343035` — the hex *of that text*, not the original bytes — and a byte-for-byte comparison against the source does not match.

### Why no behaviour changed

This is a configured default, not corruption. Two checks established that before anything was touched:

- A **keyed control table** (real primary key, same BLOB payload) renders identically — so it is independent of row identity and unrelated to the keyless work.
- The code path is an explicit `if (persist.raw.bytes) … else hex-encode`, i.e. a documented switch with `false` as its default.

So the defect is in the documentation, not the code. The old comment read:

> `persist.raw.bytes: If set to true, the connector will persist raw bytes as received in a String column.`

It describes only the `true` branch and never says what the default actually does — which is exactly why the hex output reads as data corruption to anyone comparing against MySQL. The comment now documents both modes, the precise consequence (`hex(col)` returning the hex of the text), and when to flip it — e.g. when `db_compare` checksum jobs compare binary values.

**The default is unchanged.** Flipping it would change the stored representation and require rewriting every existing table, which is not a change to make as a side effect of a documentation fix. Callers who need byte-identical binary can set `persist.raw.bytes: "true"` deliberately.

---

## Tests

`DdlReplayIdempotencyTest` (new, 5 tests) pins a guard on each generated statement, on **every clause of a multi-clause statement** — guarding only the first leaves the same defect with a narrower trigger — and on the deliberate `MODIFY` exception.

**Failing-first verified.** With the guards reverted, 4 of 5 fail, each naming the exact production error code:

```
testAddColumnIsGuarded      ... Code: 15 DUPLICATE_COLUMN
testDropColumnIsGuarded     ... Code: 10, retry loop stalls replication
testRenameColumnIsGuarded   ... Code: 10, old name no longer exists
testEveryClause...Guarded   ... expected:<2> but was:<0>
```

145 tests green overall. Pinned expectations in `MySqlDDLParserListenerImplTest` were updated for the guarded output; the MySQL **input** statements in those tests are deliberately unchanged — only the ClickHouse expectations moved.

Issue #1298, opened 2026-05-17
[BUG-DATA-6] DDL parser parseSql() return value is always null — dead code


minguyen9988:
## Summary

Both overloads of `MySQLDDLParserService.parseSql()` declare `String clickHouseResult = null` on entry and return it at the end without ever assigning it. The translated DDL is written to the `StringBuffer parsedQuery` side-channel parameter, but the return value is always null.

## Affected Code

**File:** `sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/parser/MySQLDDLParserService.java`

Lines 86-105 and 117-138:
```java
public String parseSql(String sql, String databaseName, StringBuffer parsedQuery) {
    String clickHouseResult = null;
    // ... parsing logic writes to parsedQuery ...
    return clickHouseResult;  // always null
}
```

## Impact

Any caller that relies on the return value of `parseSql()` instead of the `parsedQuery` parameter will receive null and silently lose the parsed DDL. The interface `DDLParserService` declares the return as `String`, misleading callers.

## Fix

Return `parsedQuery.toString()` instead of the never-assigned `clickHouseResult`.

## Severity

HIGH — silent data loss if callers use the return value

Issue #1297, opened 2026-05-17
[BUG-DATA-5] Non-atomic offset update — crash between DELETE and INSERT loses offset permanently


minguyen9988:
## Summary

`DebeziumJdbcStorageOperations.updateDebeziumStorageStatus()` first DELETEs the offset row, then INSERTs a new one. If the process crashes between the DELETE and the INSERT, the offset is permanently lost.

## Affected Code

**File:** `sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/parser/DebeziumJdbcStorageOperations.java`

Lines 348-350 and 395-397:
```java
// First: DELETE the existing offset
deleteOffsetStorageRow(conn, offsetKey, tableName);
// Then: INSERT the new offset  
insertOffsetStorageRow(conn, offsetKey, offsetValue, recordTimeStamp, tableName);
```

These two operations are NOT wrapped in a transaction.

## Impact

If the JVM crashes, is killed, or loses the ClickHouse connection between the DELETE and INSERT:

1. The offset row is deleted but the new one is never inserted
2. On restart, the connector has no stored offset
3. Depending on `snapshot.mode`, the connector either:
   - Re-processes the entire binlog from the beginning (massive duplicates)
   - Starts from the current binlog position (data loss for the gap)

## Fix

Since the offset table uses ReplacingMergeTree, simply INSERT the new row without deleting first. ClickHouse's ReplacingMergeTree engine will keep only the latest version after merges. This makes the operation atomic (single INSERT) and eliminates the crash window.

Alternatively, wrap the DELETE and INSERT in a transaction if ClickHouse version supports it.

## Severity

HIGH — Data loss on crash (CWE-362)

Issue #1288, opened 2026-05-17
[BUG-SEC-1] SQL Injection in DebeziumOffsetStorage — String.format constructs SQL with user-controlled topic names


minguyen9988:
## Summary

`DebeziumOffsetStorage.java` constructs SQL queries using `String.format()` with user-controlled Debezium topic names interpolated directly into the SQL string. This is a SQL injection vulnerability — a maliciously crafted topic name can execute arbitrary SQL against the ClickHouse offset storage table.

## Affected Code

**File:** `sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumOffsetStorage.java`

```java
// Line ~80: Query construction
String query = String.format("select * from %s.%s where topic_name='%s'", 
    database, tableName, topicName);

// Line ~100: Delete construction  
String deleteQuery = String.format("delete from %s.%s where topic_name='%s'",
    database, tableName, topicName);
```

The `topicName` parameter comes from Debezium's internal topic naming, which incorporates the MySQL server name from the connector config. While the server name is typically controlled by the admin, this is still a SQL injection vector:

1. If an attacker can influence the server name in the config (e.g., via the REST API or environment variable injection), they can inject arbitrary SQL
2. The pattern violates defense-in-depth — all SQL should use parameterized queries
3. ClickHouse supports parameterized queries via PreparedStatement, which is already used elsewhere in the codebase

## Fix

Replace `String.format` with a parameterized query:

```java
String query = "SELECT * FROM " + database + "." + tableName + " WHERE topic_name = ?";
PreparedStatement ps = conn.prepareStatement(query);
ps.setString(1, topicName);
ResultSet rs = ps.executeQuery();
```

(Database and table names should also be validated/escaped, but at minimum the WHERE clause value must be parameterized.)

## Impact

- **Severity:** HIGH
- **Type:** SQL Injection (CWE-89)
- **Affected versions:** All versions with DebeziumOffsetStorage

## Reproduction

Set `database.server.name` to a value containing a single quote followed by SQL, e.g.:
`test' OR 1=1; --`

The resulting query becomes:
```sql
select * from offset_db.offset_table where topic_name='test' OR 1=1; --'
```

Issue #1287, opened 2026-05-17
fix: disable.drop.truncate config is completely broken (check-before-set timing bug + unguarded CDC TRUNCATE path)


minguyen9988:
## Bug Report

The `disable.drop.truncate` config option is completely non-functional due to two independent bugs. Even with `disable.drop.truncate=true`, DROP TABLE, DROP DATABASE, and TRUNCATE TABLE from MySQL are silently forwarded to ClickHouse.

### Bug 1: DDL path timing bug (check-before-set)

In `DebeziumChangeEventCapture.performDDLOperation()`, `checkIfDDLNeedsToBeIgnored()` is called BEFORE `parseSql()`, but `parseSql()` is the method that sets the `isDropOrTruncate` flag:

```java
// Current code (broken):
StringBuffer clickHouseQuery = new StringBuffer();
AtomicBoolean isDropOrTruncate = new AtomicBoolean(false);

// BUG: isDropOrTruncate is always false here because parseSql() hasn't run yet
if (checkIfDDLNeedsToBeIgnored(DDL, props, sr, isDropOrTruncate)) {
    log.info("Ignored Source DB DDL: " + DDL);
    return;
}

// This is where isDropOrTruncate actually gets set - but the check already passed
MySQLDDLParserService mySQLDDLParserService = new MySQLDDLParserService(writer, config, databaseName);
mySQLDDLParserService.parseSql(DDL, "", clickHouseQuery, isDropOrTruncate);
```

**Fix:** Reorder so `parseSql()` runs before `checkIfDDLNeedsToBeIgnored()`.

### Bug 2: CDC data path completely unguarded

The CDC TRUNCATE path in `PreparedStatementExecutor` (triggered by Debezium CDC events with `op="T"`) has zero safety guards. Any MySQL TRUNCATE event is forwarded unconditionally through `DBMetadata.truncateTable()` regardless of the `disable.drop.truncate` setting:

```java
// Current code (no guard):
if (!truncatedRecords.isEmpty()) {
    try {
        metadata.truncateTable(conn, databaseName, tableName);
    } catch (SQLException e) {
        throw new RuntimeException(e);
    }
}
```

**Fix:** Check `disable.drop.truncate` config before calling `truncateTable()`, log a warning when blocked.

### Additional issue: Unsafe default snapshot.mode

The default `snapshot.mode` is hardcoded to `"initial"` in:
- `config.properties` (line 23)
[...]

PR #1421, opened 2026-08-27
fix: ten defects in the 2.10.1 line (TIME, NULL handling, DDL translation, identifier escaping, connection lifecycle)

minguyen9988:
Fifteen fixes to the 2.10.1 line, each with a failing-first proof and a regression test. Several were found by auditing production connector deployments rather than by reading the tracker, so they carry field evidence rather than a hypothesis.

## Issues closed by merging this PR

| Issue | Commit | Defect |
|---|---|---|
| #1215, #8 (TIME half) | `e1339a5cf` | `TIME(0)` rendered as `16:01:25.000000`; INT32 `io.debezium.time.Time` bound via `setInt`, writing raw millisecond-of-day |
| #1250 | `774da5d46` | NULL bound into a NON-nullable ClickHouse column handed to `setNull(Types.OTHER)` instead of failing |
| #335 | `ef8fe8b52` | DDL auto-create path hardcoded the delete column, ignoring `replacingmergetree.delete.column` |
| #1254 | `1c00ccdbd` | write-only success flag; an empty record list was reported as a failed batch |
| #1252 | `d55539143` | `ClickHouseBatchWriter` never closed its cached JDBC connections |
| #1298 | `8de113e43` | `parseSql()` returned an always-null local instead of the parsed query |
| #1264, #1294 | `ed7b5ab47` | database and table identifiers not escaped in generated DDL |
| #1288 | `6b599aa43` | offset-storage statements interpolated the offset key instead of binding it; the key is built from the connector name, so a quote character terminates the literal |
| #1287 | `7d512e592` | disable.drop.truncate did nothing: the DDL guard read a flag written only after it, and the CDC TRUNCATE path never checked the setting at all |
| #1297 | `ece59a296` | the offset update deleted the stored offset row before writing its replacement, so a crash in between left the connector with no offset to resume from |
| #1260 | `f97edd0aa` | the column-drop branch emitted a `delete column` clause carrying the data type -- neither the keyword nor the trailing type is accepted by the ClickHouse parser (Code: 62) |
| #1253 | `0f8d7e225` | the process-wide connection-pool cache used plain HashMap and HashSet with no synchronization anywhere in the class, and getInstance did a get-then-put that let two threads each build a pool |

## Additional fixes with no tracker issue

Found by auditing production deployments. Each is a real defect with field evidence.

| Commit | Defect |
|---|---|
| `16e6d86fb` | an untranslatable ALTER emitted a bare `ALTER TABLE <table>`, rejected `Code: 62`. Observed on 70 distinct tables across 5 deployments over 18 months |
| `cfa543eeb` | `smallint unsigned` mapped to UInt16 or Int32 depending on ZEROFILL -- 18 vs 3 columns in a single production database |
| `d0b90297b` | the default error-table name was a dotted config KEY, qualifying to `system.default.error.table`; every error the connector tried to record was discarded |
| `beecc9475` | MySQL TIME lost its sign and wrapped past 24 hours: `-01:30:00` became `22:30:00`, `838:59:59` became `22:59:59` |
| `02a133495` | a table-level `PRIMARY KEY (a,b)` emitted `Nullable(T)` columns while naming them in ORDER BY -- DDL ClickHouse rejects with `Code: 44` |

Two of those -- the TIME wrap and the bare ALTER -- are silent: wrong or missing data with no exception raised.

## Verification

```
sink-connector                     321 tests   0 failures   0 errors   2 skipped
sink-connector-lightweight (DDL)   160 tests   0 failures   0 errors   1 skipped
sink-connector-lightweight (offset) 10 tests   0 failures   0 errors   0 skipped
```

Every fix has a failing-first proof: the test was written, run before the change to capture the verbatim failure, then re-run after. Control assertions accompany each one so existing behaviour is pinned rather than assumed.

The image built from an earlier head of this branch was verified end to end against a live MySQL to ClickHouse pipeline -- snapshot plus CDC -- covering TIME rendering at three precisions, the signed and large-range TIME values, table-level PRIMARY KEY auto-create, unsigned determinism, untranslatable ALTERs, the configurable delete column with a non-default name, and a DOUBLE/TINYINT regression sweep. Zero exceptions in the connector log.

## Deliberately out of scope

- `ClickHouseAlterTable` emits an unqualified table name. `ed7b5ab47` escapes it but does not add a database qualifier -- that is a behavioural change of a different kind and deserves its own commit and proof.
- A keyless source table still auto-creates as ReplacingMergeTree with `ORDER BY tuple()`, collapsing rows. Both alternatives have already been tried and abandoned in this codebase with measurements behind the decision: ordering by all columns freezes the schema (Code 524/47), and a row hash makes columns undroppable (Code 44). Changing it would also put the connector at odds with keyless tables already created in the field. It needs its own PR and its own discussion.
- One pre-existing test failure, `DataTypeConverterTest.testAllDataTypeConversions` (`FLOAT expected Float32, got Float64`), fails at the base of this branch both before and after these changes. Verified against a clean checkout of the tip -- not introduced here, and not fixed here.


minguyen9988:
Added a real fix on top of the docs commit, so the image this PR builds is now worth testing rather than just a rebuild of `develop`.

## Two MySQL TIME defects, both silent

**1. Spurious sub-second digits — closes #1215, and the TIME half of #8**

`DebeziumConverter.MicroTimeConverter` formatted with a fixed-width pattern:

```java
time.format(DateTimeFormatter.ofPattern("HH:mm:ss.SSSSSS"))
```

so a TIME with no fractional component rendered as `16:01:25.000000`. The converter got no precision or target-type context, unlike `MicroTimestampConverter` right beside it, which already branches on the ClickHouse type.

**2. Raw millisecond-of-day written for INT32 TIME — not previously filed**

`ClickHouseDataTypeMapper` maps `(INT32, io.debezium.time.Time)` to a String column, but `isFieldTime` was set only for `INT64 + MicroTime`. An INT32 `Time` matched no time branch, fell through to the integer path, and was bound by `ps.setInt` — writing `57685000` instead of `16:01:25`.

This is the shape Debezium emits under the **default** `ADAPTIVE` precision mode for `TIME(0)`..`TIME(3)` (`TemporalPrecisionMode.ADAPTIVE`: `p<=3` to `Time`/INT32, `p<=6` to `MicroTime`/INT64), so it affects the common configuration rather than an edge case.

## Fix

Rendering emits only the fractional digits actually present, via `DateTimeFormatterBuilder.appendFraction(NANO_OF_SECOND, 0, 6, true)`. A `TimeConverter` is added for the INT32 milliseconds-past-midnight shape, and the mapper routes INT32 + `Time.SCHEMA_NAME` into the time branch, dispatching INT32 to `TimeConverter` and INT64 to `MicroTimeConverter`.

On trimming trailing zeros: the Debezium schema for a TIME column carries no scale, so `TIME(0) '16:01:25'` and `TIME(6) '16:01:25.000000'` arrive as the identical wire value — no rendering can reproduce both source texts. Trimming never discards value information; padding asserts a microsecond scale for which there is no evidence, and is wrong for the default case. The existing `removeTrailingZeros` helper was deliberately not reused: it strips zeros from the whole string, turning `16:01:20` into `16:01:2`.

## Failing-first evidence

Before the fix:

```
ClickHouseDataTypeMapperTimeTest.int32TimeIsBoundAsTimeString:120
  a TIME value must be bound as a string, not as an integer; setInt bound
  the raw millisecond-of-day count 57685000
  ==> expected: <setString> but was: <setInt>
ClickHouseDataTypeMapperTimeTest.int64MicroTimeIsBoundAsTimeString:158
  expected: <16:01:25> but was: <16:01:25.000000>
DebeziumConverterTest.testMicroTimeConverterWholeSecondHasNoFraction:318
  expected:<16:01:25[]> but was:<16:01:25[.000000]>
DebeziumConverterTest.testMicroTimeConverterMillisecondsAreNotPadded:334
  expected:<17:51:04.777[]> but was:<17:51:04.777[000]>
```

All pass after the fix, with control assertions covering genuine sub-second values, midnight, and the INT64 path.

One pre-existing assertion encoded the padded output (`09:01:01.000000`). It was updated rather than deleted, with the reason inline, so the behaviour change stays visible in the diff.

```
module suite   270 tests   0 failures   0 errors      (baseline 262/0/0)
```

The delta is exactly the new tests.

## Verification of the previous image from this PR

The image built from the prior head was pulled and exercised against a live MySQL to ClickHouse pipeline. All eight merged fixes were confirmed present in the shipped jar's bytecode, and four were exercised end to end: DOUBLE auto-creating as `Float64` with exact high-precision round-trip, TINYINT signed/unsigned boundaries, `_version` never binding UInt64 max, and `/metrics` returning `text/plain; version=0.0.4; charset=utf-8`. Zero exceptions in the connector log across snapshot and CDC.

One environment note for anyone testing the image: it has no `/etc/ssl/certs` directory, so on hosts whose container runtime bind-mounts CA bundles into every container the runtime fails before Java starts. Unrelated to these changes, but worth knowing.

minguyen9988:
### Closes #1250 — NULL handling crash in non-nullable ClickHouse columns

Commit `774da5d46`.

**What was wrong.** `PreparedStatementFieldMapper.insertPreparedStatement` had two sites that handed SQL NULL to the driver without looking at the target column:

- `PreparedStatementFieldMapper.java:237-240` — `if (value == null) { ps.setNull(index, Types.OTHER); continue; }`
- `PreparedStatementFieldMapper.java:250-255` — the `DataException` catch logged *"Setting column %s to NULL might fail for non-nullable columns"* and then called `ps.setNull(index, Types.OTHER)` anyway

with the matching admission at line 226: `//ToDO: Setting null to a non-nullable field) will throw an error.`

**The nullability was already there.** `DBMetadata.getColumnsDataTypesForTable` (`DBMetadata.java:509-568`) runs `SELECT name, type, default_kind FROM system.columns ...` and returns the verbatim ClickHouse type strings. That map is passed straight into `insertPreparedStatement` as `columnNameToDataTypeMap` (`PreparedStatementFieldMapper.java:173`) and is the very map the bind loop iterates at line 177 — so `Nullable(String)` vs `String` was distinguishable one line above the `setNull` call. Nothing had to be fetched; the information was ignored. No refactor was needed.

**The fix.** `acceptsNull(String)` classifies the declared type and `setNullChecked(...)` replaces both bare `setNull` sites. Nullable columns are untouched — they reach `ps.setNull(index, Types.OTHER)` byte-for-byte as before.

**Why fail loudly instead of substituting a default.** The issue suggests coercing to a type-appropriate default (`0`, `''`, epoch). That is the one outcome nobody can detect afterwards: it writes a row that never existed in the source while the row count still matches, so count-based checksums report the table clean. The status quo is barely better — the driver fails the batch with an error naming neither the column nor the table, leaving the operator to bisect thousands of records. So the batch is failed at the bind site with database, table, column and declared type in the message. Failing is recoverable (retry, or a one-line `ALTER`); a silently coerced row is not. Same reasoning and same exception type as the adjacent `rejectUnderivableVersion` (#1213). Nothing is muted — the existing ERROR log on the `DataException` path is kept, with its wording corrected to say what now happens.

Type classification details:

- `LowCardinality(...)` is a storage wrapper, so it is unwrapped: `LowCardinality(Nullable(String))` accepts NULL, `LowCardinality(String)` does not.
- `Array(Nullable(String))` does **not** accept NULL — the array itself is required, only its elements may be NULL. A "contains Nullable" match would have waved it through; the check is anchored at the start.
- An unknown or blank type returns `true`. An undeterminable type is not proof that a column rejects NULL, and a guess must never fail a batch that works today.
- Columns the connector populates itself (`_version`, `_sign`, `is_deleted`, the replication-history columns, Kafka metadata, the raw-data column) are exempt: they are deliberately placeholder-nulled and overwritten by the handlers further down, and are non-nullable in the schemas the connector generates.

**Failing-first evidence.** New `PreparedStatementFieldMapperNullabilityTest` (8 tests) drives the real bind loop through a recording JDBC proxy (Mockito is not on the test classpath). Run against the **unpatched** mapper — the two defect tests fail, the three control tests already pass:

```
Tests run: 5, Failures: 2, Errors: 0, Skipped: 0

nullIntoNonNullableColumnIsRejected:145 a NULL bound into a non-nullable String
  column must fail the batch here, not be handed to the driver as
  setNull(Types.OTHER) ==> Expected java.lang.IllegalStateException to be
  thrown, but nothing was thrown.

columnAbsentFromSourceIsRejectedForNonNullableTarget:180 a ClickHouse column
  absent from the source record must not be force-nulled when the column cannot
  store NULL ==> Expected java.lang.IllegalStateException to be thrown, but
  nothing was thrown.
```

After the fix:

```
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

Controls pinning unchanged behaviour: `Nullable(String)` still binds `setNull` at its own index; `LowCardinality(Nullable(String))` still binds; the connector-populated `_version` column keeps its placeholder NULL.

**Full module suite.**

| | tests | failures | errors | skipped |
|---|---|---|---|---|
| baseline (this branch, before) | 270 | 0 | 0 | 2 |
| after | 278 | 0 | 0 | 2 |

Delta is exactly the 8 tests added here.

minguyen9988:
Added a third fix to this PR: the DDL auto-create path ignoring `replacingmergetree.delete.column`. Closes #335.

## Defect

The record-schema auto-create path honours the configured delete column (`ClickHouseAutoCreateTable.java:160-163`, fed by `DbWriter.java:185-199`), but the lightweight DDL path hardcoded it:

```java
// MySqlDDLParserListenerImpl.java:390
String isDeletedColumn = IS_DELETED_COLUMN;
```

used for the column at `:430` and for the engine clause at `:445`/`:447`. A table auto-created by replicating MySQL DDL therefore always got `is_deleted`, whatever the user configured.

The collision-rename at `:398` had the same flaw — it renamed a clashing source column to `_is_deleted` even when the configured name was different, so the rename could miss the column it was meant to protect.

## Fix

`resolveDeleteColumn()` reads the configured value; the column, the engine clause and the collision rename all use the resolved name. No signature change was needed — `this.config` is already a field (`:93`, assigned `:146`) and `originalsStrings()` was already used twice in the file (`:439`, `:1022`).

## On the default — #243 deliberately not changed

The distinction that matters is `originalsStrings()` versus `getString()`. `getString()` would substitute the ConfigDef default `"sign"` (`ClickHouseSinkConnectorConfig.java:411-413`) — a third spelling, matching neither the `is_deleted` this path has always emitted nor the `_sign` of Kafka-era tables. That would silently change the delete column for every deployment that never set the key, so a table created tomorrow would differ from one created yesterday.

Reading the originals keeps "user set nothing" distinguishable from "user asked for the default", so behaviour is unchanged unless the key is explicitly set. A control test pins that. Standardising the default belongs in a major release with a migration note, not in a bug fix.

## Pre-existing limitation found while checking impact — not addressed here

`QueryFormatter` (`:527`, `:547`, `:565`, `:623`, `:655`, `:684`, `:714`) and `PreparedStatementFieldMapper.isUnboundByDesign` (`:56-59`) match the delete column against the hardcoded constants rather than the configured name, including a literal `" AND \`is_deleted\` = 0"`. A custom name is therefore still not honoured end to end in the replication-history path. This change neither creates nor worsens that, but it deserves its own issue.

The write path is safe: `DbWriter.configureReplacingMergeTreeColumns` (`:344-355`) parses the delete column out of the existing table's actual `ReplacingMergeTree(_version, X)` clause and only falls back to config when the clause has no second element, so it follows whatever the DDL emitted.

## Failing-first evidence

Tests applied without the fix:

```
testCreateTableHonorsConfiguredDeleteColumn:2681
  expected:<...`[_sign` UInt8) Engine=ReplacingMergeTree(_version,_sign])...>
   but was:<...`[is_deleted` UInt8) Engine=ReplacingMergeTree(_version,is_deleted])...>
testConfiguredDeleteColumnCollisionRename:2757
  expected:<...`[__sign` UInt8) Engine=ReplacingMergeTree(_version,__sign])...>
   but was:<...`[is_deleted` UInt8) Engine=ReplacingMergeTree(_version,is_deleted])...>
testReplicatedEngineHonorsConfiguredDeleteColumn:2781
  expected:<...`[_sign` UInt8) Engine=ReplicatedReplacingMergeTree(_version, _sign])...>
   but was:<...`[is_deleted` UInt8) Engine=ReplicatedReplacingMergeTree(_version, is_deleted])...>
```

Two control tests (unset key, blank value) passed before the fix, so they pin existing behaviour rather than the change. After the fix: 113 run, 0 failures. `sink-connector` module: 278 tests, 0 failures, 0 errors.

## PR state

This branch now carries four commits: the docs note, the MySQL TIME fix (#1215 and the TIME half of #8), the non-nullable NULL rejection (#1250), and this one (#335).

The image built from the previous head was verified against a live MySQL to ClickHouse pipeline — the TIME assertions all passed (`TIME(0)` renders `16:01:25` rather than `16:01:25.000000` or a raw integer; `TIME(3)` and `TIME(6)` keep their fractions), with a clean regression sweep and zero exceptions in the connector log.

minguyen9988:
Two more fixes on this PR. Closes #1252 and #1254.

## #1252 — JDBC connections never closed

`ClickHouseBatchWriter` cached a connection per database (`:55`, populated at `:144-158`) and never closed any of them; the class had no `close()` and did not implement `AutoCloseable`. `systemConnection()` (`:465-470`) additionally replaced an unusable handle without closing the old one, leaking one more per reconnect.

The amplifier is the in-process restart path: `ClickHouseDebeziumEmbeddedApplication:245-249` to `DebeziumChangeEventCapture.java:1440` builds a fresh writer on every engine restart, so the leak scales with restart count rather than being bounded.

Fix: `implements AutoCloseable` with an idempotent, non-throwing `close()` that closes every cached connection and the system connection; `systemConnection()` closes the stale handle before replacing it; and — the part that actually matters — the lifecycle is **invoked** from `DebeziumChangeEventCapture.stop()` after engine and executor termination. A `close()` nobody calls is not a fix, so there is a test that drives the real `stop()` rather than calling `close()` directly.

Failing-first, writer half:

```
Tests run: 3, Failures: 2
  no lifecycle method exists to invoke
  the stale system connection was replaced WITHOUT being closed
    expected:<1> but was:<0>
```

Failing-first, owner half (drives the real `stop()`):

```
Tests run: 3, Failures: 2
  expected:<3> but was:<0>
```

After: `ClickHouseBatchWriterConnectionLifecycleTest` 8/8, `SingleThreadedWriterShutdownTest` 3/3.

## #1254 — write-only success flag

The reported masking scenario **cannot occur at head**, and saying so plainly matters more than claiming a fix: every failure path rethrows (`:358-363`), so `return result.get()` at `:368` is unreachable on failure.

The flag was wrong in the other direction. An empty record list runs zero `forEach` iterations (`:194`), keeps the initial `false` (`:190`), and reports "nothing executed" to `addToPreparedStatementBatch:149-157` as "batch failed" — which the #1285 change now escalates into a `BatchPersistenceException`. So the flag turned a no-op into a spurious failure.

Fix: removed the flag; `executePreparedStatement` returns `void` and the exception that already governs control flow is the single source of truth. Rethrow semantics are untouched, since #1285 depends on failures reaching the caller.

Failing-first:

```
Tests run: 5, Failures: 1
  emptyEntryIsNotReportedAsFailure ... <<< FAILURE!
```

After: `Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`.

## Suite

```
sink-connector   291 tests   0 failures   0 errors   2 skipped     (baseline 278/0/0/2)
```

Delta is exactly the new tests.

## PR state

Six commits now: the docs note, MySQL TIME (#1215 and the TIME half of #8), non-nullable NULL rejection (#1250), the DDL delete-column fix (#335), and these two.

minguyen9988:
Added a fix for a DDL-translation defect found by auditing production connector logs rather than by reading code. This one has real mileage behind it: **70 distinct tables across 5 deployments over 18 months.**

## Defect

`MySqlDDLParserListenerImpl.enterAlterTable()` appends the statement prefix unconditionally the moment it sees the `TableNameContext`:

```java
// :1331 / :1333, Constants.ALTER_TABLE = "ALTER TABLE %s"
this.query.append(String.format(Constants.ALTER_TABLE, databaseName + "." + this.tableName));
```

It then walks the sibling clauses and appends only for recognised context types. When none of them is recognised, what reaches ClickHouse is the prefix alone:

```
Code: 62. DB::Exception: Syntax error: failed at position 32 (end of query)
```

Shapes observed doing this in production:

```
ALTER TABLE t ADD CONSTRAINT `x` FOREIGN KEY (c) REFERENCES o(id)
ALTER TABLE t DROP FOREIGN KEY `x`
ALTER TABLE t ADD INDEX i (col), ALGORITHM=INPLACE, LOCK=NONE
ALTER TABLE t ALTER INDEX i INVISIBLE
```

One production case emitted a statement ending in a bare trailing comma, which is the tell that clauses are being stripped textually rather than the statement being rebuilt from the AST.

The practical impact is not the failed statement — it is that a schema change silently does not happen, and the operator has only a `Code: 62` in a log to notice it by.

## Fix

`exitAlterTable()` takes the prefix back when nothing followed it.

Two details worth calling out, because they are what makes this safe:

1. The emptiness test is on **what actually landed in the buffer**, not on which clause types were recognised. That distinction preserves `ADD CHECK CONSTRAINT`, which appends from a separate listener callback *after* the walk finishes — a recognition-based test would have wrongly discarded it.
2. The prefix is recorded as an offset **plus its exact text**, and compared before deleting. So `ALTER ... RENAME TO`, which rewrites the buffer wholesale, makes this a no-op rather than a truncation.

A skipped statement is logged at WARN naming the table, so an untranslatable schema change is visible rather than silent.

Mixed statements are unaffected: `ALTER TABLE t ADD COLUMN c INT, ADD INDEX i (c)` still applies the column half and drops the index half, exactly as before.

## Failing-first evidence

15 tests, **5 failures / 0 errors** before the fix — all four production shapes emitting the bare prefix:

```
expected:<[]> but was:<[ALTER TABLE employees.line_items]>
```

After: **15 / 0 / 0**. The 10 control tests passed both before and after, so they pin existing behaviour rather than the change. Targeted run including the full existing DDL parser suite: **128 tests, 0 failures, 0 errors**.

One correction made during the work rather than papered over: the first `ADD INDEX` repro errored on a parse failure because the comma MySQL requires before `ALGORITHM=` was missing. That was a broken test, not a reproduction — it was fixed and re-run to get a clean 5-failure proof.

## Related, investigated but deliberately not shipped here

A keyless source table auto-creates as ReplacingMergeTree with `ORDER BY tuple()`, which collapses every row into one. There are two independent `tuple()` paths — `ClickHouseAutoCreateTable.java:265` (record-driven, completely silent) and `MySqlDDLParserListenerImpl.java:540` (DDL-driven, which does log a banner) — and `KeylessTablePreflight` covers neither of them in the record-driven case.

I am not changing the engine or the synthesised sorting key in this PR. Both alternatives have already been tried and abandoned in this codebase with measurements behind the decision: ordering by all columns freezes the schema (Code 524/47), and a row hash makes columns undroppable (Code 44). Changing it now would also put the connector at odds with every keyless table already created in the field. That deserves its own PR and its own discussion.

Two smaller things surfaced while investigating it, worth recording: the record-driven path emits no warning at all for a keyless table, and an *empty* (non-null) primary key list produces `PRIMARY KEY() ORDER BY()` — invalid SQL — because the presence check returns vacuously true for a list naming no columns.

minguyen9988:
Two more fixes, both found by auditing production deployments rather than by reading code.

## Unsigned integer mapping was not deterministic

A full 291-column diff of one production database against its ClickHouse replica showed MySQL `smallint unsigned` resolving to **UInt16 for 18 columns but Int32 for 3** — same database, same connector.

Root cause at `DataTypeConverter.java:136`: the unsigned override was looked up under the *raw* resolved type name. MySQL implies UNSIGNED from ZEROFILL, and Debezium folds both into the name, so the lookup key becomes `"SMALLINT UNSIGNED ZEROFILL"` — not a map entry — and falls through to the signed mapping. Confirmed by driving the parser directly:

```
SMALLINT UNSIGNED          -> "SMALLINT UNSIGNED"          -> UInt16
SMALLINT UNSIGNED ZEROFILL -> "SMALLINT UNSIGNED ZEROFILL" -> Int32
```

That is exactly the observed 18-vs-3 split. The unsignedness is available on both paths, so this is a genuine defect rather than an information gap. The same probe found every ZEROFILL width affected, and aliases wrong in the other direction — `INT1 UNSIGNED` resolving to UInt32 through a prefix fallback when it should be UInt8.

Fix: `normalizeMysqlTypeName()` strips display width, ZEROFILL and SIGNED before lookup; the prefix fallback is replaced by an exact match; alias keys are added. Standardised on **UInt16**, the faithful mapping for `0..65535` and already the majority behaviour. This affects CREATE only — no existing column changes type, so there is no migration implied.

```
DataTypeConverterTest.testUnsignedConversionsAreDeterministic:254
  Failed for test case: SMALLINT UNSIGNED ZEROFILL expected:<[UInt16]> but was:<[Int32]>
ClickHouseDataTypeMapperTest:97
  smallint unsigned must always map to UInt16: INT2 UNSIGNED expected:<UInt[16]> but was:<UInt[32]>
```

Worth flagging: `DataTypeConverterTest.testAllDataTypeConversions` fails on this branch **both before and after** this change (`FLOAT expected Float32, got Float64`). Verified against a clean checkout of the tip — pre-existing, not introduced here. The new assertions were put in their own method so the two cannot be confused.

## The error table never received anything

The name the connector qualifies is not a valid identifier, and it reproduces from an **empty config** — a code default, not operator misconfiguration:

```
ConfigVariables.java:96                  ERROR_TABLE_NAME("default.error.table")
ClickHouseSinkConnectorConfig.java:747   passes the config KEY as the ConfigDef DEFAULT VALUE
ErrorLogger.java:46-61 / :93-106         qualifies it as "system" + "." + <value>
```

The emitted identifier is `system.default.error.table` — four dotted parts. ClickHouse cannot resolve it, so every error the connector tries to record is discarded.

This also explains why the errors table is absent in the field: the `replica_source_error` fallback at `ErrorLogger.java:23` is **dead code**. The config always supplies a non-empty value, so the `isEmpty()` branch never fires and the intended default is never used.

The consequence is not a noisy log line. Failures that are supposed to leave a durable record leave none — which compounds any defect whose only other trace is a log entry, including the untranslatable-ALTER case fixed earlier in this PR.

Fix: the default is the bare table name, and the qualified name is validated before use — a value that does not resolve to a single `database.table` pair fails loudly naming what it received, instead of silently issuing SQL that cannot work.

```
ErrorLoggerTableNameTest.defaultErrorTableNameIsABareIdentifier:20
java.lang.AssertionError: The default error table name must not contain a dot;
a dotted value is concatenated into an unusable multi-part identifier such as
system.default.error.table. Got: default.error.table
```

The test is constructed from an empty config so it reproduces the shipped default rather than a hand-made string.

## Suite

```
sink-connector   297 tests   0 failures   0 errors   2 skipped     (baseline 291/0/0/2)
```

Delta is exactly the new tests.

minguyen9988:
## Two more estate-found defects: signed/large MySQL TIME, and a nullable PRIMARY KEY column in the sorting key

Both were found by scanning replicated table shapes in a production estate against what the connector actually emits. Each is a separate commit with a failing-first proof.

---

### 1. MySQL TIME loses its sign and wraps past 24 hours — `beecc947`

**Status of the reported symptom.** The estate finding was that the same logical MySQL TIME arrives in ClickHouse in two different string shapes — `19:00:00.000000` from a `time(6)` column and `23:23:00` from a `time` column. That half is **already fixed** earlier in this branch (`e1339a5c`): rendering now emits only the fractional digits actually present, so both render as `19:00:00` / `23:23:00` and sort together. The control test in this commit pins it. No further work was needed there, and none was invented.

**But the signed-range half is real, and worse.** MySQL `TIME` is a signed *duration* with a declared range of `-838:59:59 .. 838:59:59` — not a time of day. Both converters rendered it through `java.time.LocalTime`, which can represent neither end of that range.

Root cause — `DebeziumConverter.java:73` (`MicroTimeConverter.convert`) and `:98` (`TimeConverter.convert`), both doing:

```java
Instant.EPOCH.plus(value, MICROS|MILLIS).atZone(UTC).toLocalTime()
```

`Instant.plus` accepts a negative or over-a-day amount happily, and `toLocalTime()` then takes the time-of-day of the resulting instant. Measured on the tip before the fix:

| MySQL value | wire value | replicated as |
|---|---|---|
| `-01:30:00` | `-5400000000` micros | `22:30:00` |
| `838:59:59` | `3020399000000` micros | `22:59:59` |
| `-01:30:00` | `-5400000` millis | `22:30:00` |
| `30:00:00` | `108000000` millis | `06:00:00` |

No exception, no log line. The destination looks healthy; only a checksum comparison against the source would ever reveal it.

**These values genuinely reach the converter.** `JdbcValueConverters` calls `MicroTime.toMicroOfDay(value, acceptLargeValues)` / `Time.toMilliOfDay(value, acceptLargeValues)` with `acceptLargeValues = supportsLargeTimeValues()`, which returns true for `ADAPTIVE`, `ADAPTIVE_TIME_MICROSECONDS`, `MICROSECONDS` and `NANOSECONDS` — every temporal precision mode this connector runs under. With that flag set Debezium skips its own `isNegative()`/`ONE_DAY` range check and returns the raw `Duration` verbatim. Verified by disassembling `debezium-core` 3.1.3.Final.

**Fix.** `renderSignedTime()` renders straight from the duration when the value does not fit a clock: sign leading the whole value (MySQL's own `-01:30:00.5`, never `-01:-30:-00.5`), hour field not truncated to two digits, and the same fraction rule as the existing formatter. In-range values keep the `LocalTime` path and render byte-for-byte as before.

Failing first, at tip `d0b90297b`, before touching `DebeziumConverter.java`:

```
Tests run: 24, Failures: 5, Errors: 0, Skipped: 0

DebeziumConverterTest.testMicroTimeConverterBeyondTwentyFourHours:405
    expected:<[838]:59:59> but was:<[22]:59:59>
DebeziumConverterTest.testMicroTimeConverterNegativeValue:394
    expected:<[-01]:30:00> but was:<[22]:30:00>
DebeziumConverterTest.testMicroTimeConverterNegativeWithFraction:419
    expected:<[-00:00:00].5> but was:<[23:59:59].5>
DebeziumConverterTest.testTimeConverterBeyondTwentyFourHours:458
    expected:<[30]:00:00> but was:<[06]:00:00>
DebeziumConverterTest.testTimeConverterNegativeValue:445
    expected:<[-01]:30:00> but was:<[22]:30:00>
```

The control test passed in that same pre-fix run, so the five failures are the defect and not a broken harness. After the fix: `Tests run: 24, Failures: 0`.

One pre-existing test was **corrected rather than accommodated**. `testMicroTimeConverter` built its input as `ZonedDateTime.of(...).toEpochSecond() * 1000 * 1000` = `1704099661000000` micros — an epoch timestamp, ~473361 hours past midnight. That is not a time of day, no MySQL TIME can hold it, and Debezium never emits it for a TIME column; it only read as `09:01:01` because the wrap bug folded it onto a clock. The input is now `09:01:01` as micros past midnight; the assertion and what it checks (a whole-second TIME renders with no fraction) are unchanged.

Not fixed, and out of this connector's hands: Debezium truncates `toMilliOfDay` to `int`, so a `TIME(0)..TIME(3)` value past ~24.8 days overflows before the connector sees it. MySQL's own range caps well below that, so no legal value is affected.

---

### 2. A table-level PRIMARY KEY column is emitted `Nullable` and lands in `ORDER BY` — `02a13349`

The estate finding was tables with a nullable column in the `ORDER BY` of a ReplacingMergeTree, one of them worked around by hand as `ifNull(col,'')`. Tracing where such DDL could come from turned up a concrete, reachable bug on the lightweight auto-create path.

MySQL forces every PRIMARY KEY column to `NOT NULL`, silently, whether or not the column declaration says so. Verified on MySQL 8.0.36:

```
CREATE TABLE t1 (id INT AUTO_INCREMENT, x INT, PRIMARY KEY (id));
SHOW COLUMNS FROM t1;   ->  id  int  Null=NO  Key=PRI
```

The parser did not apply that rule, and the column it did not apply it to is the one it then puts in the sorting key. In `MySqlDDLParserListenerImpl.java`:

- `:928` `parseColumnDefinitions` — `isNullColumn` is cleared only by an explicit `NULL`/`NOT NULL` constraint or by a **column-level** `PRIMARY KEY`.
- `:791` `parseCreateTable` — a **table-level** `PRIMARY KEY (...)` is parsed in a different branch entirely, which appends the columns to `orderByColumns` and never revisits their nullability.
- `:985` the column is therefore emitted `Nullable(T)` …
- `:553` … and the same column is emitted as the `ORDER BY`.

This is not a subtle semantic worry — the DDL does not execute. Measured on ClickHouse 24.8.14.10547:

```
CREATE TABLE nk_probe(id Nullable(Int32), x Nullable(Int32),
  _version UInt64, is_deleted UInt8)
  Engine=ReplacingMergeTree(_version,is_deleted) ORDER BY (id)

Code: 44. DB::Exception: Sorting key contains nullable columns, but merge
tree setting `allow_nullable_key` is disabled. (ILLEGAL_COLUMN)
```

The table is never created and replication of it cannot start. `allow_nullable_key` is emitted only for the all-columns keyless fallback (`:590`), deliberately and correctly — it is not emitted here, and adding it would be the wrong fix.

**Fix, and why this shape.** A pre-pass records the table-level PRIMARY KEY columns before any column is emitted (necessary because the query is built streaming while the `PRIMARY KEY` clause comes after the columns it names), and the column pass clears `isNullColumn` for them.

The `ORDER BY` clause is **byte-for-byte unchanged** — only the column *type* changes, from `Nullable(T)` to `T NOT NULL`, which is what the source column has always been. This matters for the tables already in the field: a sorting key cannot be altered after creation, so a fix that rewrote the `ORDER BY` (wrapping as `ifNull`/`assumeNotNull`, as the production workaround does) could not be adopted by an existing table without a rebuild, and would enshrine a guard against a value the source cannot produce.

It is also consistent with the decision this codebase already makes one step away: a UNIQUE key spanning nullable columns is **refused** as a sorting key (`:394`) because MySQL does not treat NULLs as equal there, so such an index is genuinely not a row identity. A PRIMARY KEY is the opposite case — MySQL guarantees `NOT NULL` — so the consistent action is to record the guarantee. The UNIQUE path is deliberately untouched.

The other auto-create path, `ClickHouseAutoCreateTable.java:249`, does **not** have this defect: its nullability comes from `Field.schema().isOptional()` (`:145`) and Debezium marks key fields non-optional, so its key columns are already `NOT NULL`. Untouched.

Failing first, before touching the parser:

```
Tests run: 6, Failures: 3, Errors: 0, Skipped: 0

testTableLevelPrimaryKeyColumnIsNotNullable:83
  Was: CREATE TABLE if not exists employees.pk_tbl(id Nullable(Int32),
  x Nullable(Int32),`_version` UInt64,`is_deleted` UInt8)
  Engine=ReplacingMergeTree(_version,is_deleted) ORDER BY (id)

testCompositeTableLevelPrimaryKeyColumnsAreNotNullable:113
  a was emitted Nullable, was: ... employees.pk_comp(a Nullable(Int32),
  b Nullable(Int32),v Nullable(String), ...) ORDER BY (a,b)

testQuotedTableLevelPrimaryKeyColumnIsNotNullable:141
  emp_no was emitted Nullable, was: ... employees.`pk_q`(`emp_no`
  Nullable(Int32),`note` Nullable(String), ...) ORDER BY (`emp_no`)
```

All three control tests — explicit `NOT NULL` PRIMARY KEY, column-level PRIMARY KEY, and a table with no PRIMARY KEY at all — passed in that same pre-fix run. After the fix: `Tests run: 6, Failures: 0`.

Three pre-existing expectations were **corrected, not accommodated**: `testCreateTableAutoIncrement`, `testCreateTableWithPrimaryKey` and `testCreateTableWithRangePartitionByYearFunction` each pinned the exact DDL string ClickHouse rejects with Code: 44. They now expect the key column `NOT NULL`; every non-key column stays `Nullable` and every `ORDER BY` is unchanged.

---

### Suite results

Runs / failures / errors / skipped.

| Suite | Baseline | After | Delta |
|---|---|---|---|
| `sink-connector` ALL | 297 / 0 / 0 / 2 | 303 / 0 / 0 / 2 | +6, exactly the new TIME tests (5 defect + 1 control) |
| `sink-connector-lightweight` (targeted DDL-parser set) | 154 / 0 / 0 / 1 | 160 / 0 / 0 / 1 | +6, exactly the new `CreateTablePrimaryKeyNotNullTest` |

Zero failures on both, and the delta on each is only the tests added here.

`DataTypeConverterTest.testAllDataTypeConversions` is a known pre-existing failure at tip (FLOAT expected Float32, got Float64), unrelated to these changes and not touched.

minguyen9988:
Two more fixes. Closes #1298, #1264 and #1294.

## #1298 -- `parseSql()` returned a value that was always null

Both overloads in `MySQLDDLParserService` declared `String clickHouseResult = null;` (`:88`, `:118`) and returned it unassigned (`:104`, `:137`). A grep for `clickHouseResult` returns exactly those four lines, so there was provably no assignment in between. The real output goes into a `StringBuffer` side-channel passed by the caller.

Caller audit before changing the signature: two production callers -- `DebeziumEmbeddedRestApi.java:326` and `DebeziumChangeEventCapture.java:769` -- both discard the return and read the buffer, and no test caller assigns it either. So nothing depended on the null.

Fix: return `parsedQuery.toString()`, restoring the contract the interface already documents at `DDLParserService.java:23,36`. Removing the return was the alternative, but the documented contract is the lower-risk restoration and leaves no trap for a future caller.

```
ParseSqlReturnValueTest.testThreeArgOverloadReturnsTranslation:66
  parseSql documents a String return but handed back null
ParseSqlReturnValueTest.testFourArgOverloadReturnsTranslation:98
  parseSql documents a String return but handed back null
ParseSqlReturnValueTest.testUntranslatableStatementStaysEmpty:157
  expected:<> but was:<null>

Tests run: 5, Failures: 4, Errors: 0
```

Four rather than five because the CONTROL passed -- which is itself the evidence that the control's expected strings are byte-exact against the unmodified parser. After: 5/5.

## #1264 and #1294 -- identifiers not escaped in generated DDL

A database or table whose name is a reserved keyword, or contains a hyphen or space, produces DDL the server rejects. Six sites, one defect class:

```
ClickHouseAutoCreateTable.java:126-128   escapes the table name, appends the DATABASE name bare
ClickHouseAutoCreateTable.java:342       "CREATE DATABASE IF NOT EXISTS " + databaseName   (#1294)
ClickHouseAlterTable.java:73-74          table name appended unescaped
ClickHouseCreateDatabase.java:28         String.format with a bare %s for dbName
ClickHouseBatchRunnable.java:253         "CREATE DATABASE IF NOT EXISTS " + databaseName
```

Measured against ClickHouse 24.8 rather than assumed:

```
CREATE DATABASE IF NOT EXISTS my-db     Code: 62 SYNTAX_ERROR at position 33 ('-')
CREATE TABLE my-db.`t1`                 Code: 62 at position 16 ('-')
ALTER TABLE my-tbl add column ...       Code: 62 at position 15 ('-')
all three quoted                        rc=0
```

Fix: one `quoteIdentifier` static helper on `ClickHouseTableOperationsBase` -- the shared parent of AutoCreateTable, AlterTable and CreateDatabase -- used at every site. BatchRunnable does not extend that class but calls the public static, so the rule lives in exactly one place. No existing helper was reused because none exists: a repo-wide grep found nothing, and `ClickHouseUtils.escape` from clickhouse-data is backslash escaping (decompiled to confirm), which is the wrong tool for identifiers.

Escape rule is backtick-wrap plus doubling any embedded backtick, verified live: doubling yields the literal name rather than terminating the identifier.

**Deliberately out of scope.** `ClickHouseAlterTable` emits an *unqualified* table name, relying on the connection's default database. This change escapes that name but does not add a database qualifier -- qualification is a behavioural change of a different kind, not covered by the evidence above, and deserves its own commit and proof.

**Existing deployments are unaffected.** Backticking a normal identifier does not change resolution: plain `order` and quoted `order` both return rc=0. A control test pins that a plain identifier still yields a parseable qualified name.

Five existing expectations changed -- three in `ClickHouseAutoCreateTableTest`, two in `ClickHouseAlterTableTest` -- all for the same reason and all backticks-only. No other expectation was touched.

```
IdentifierQuotingTest.testCreateTableQuotesDatabaseName:94
  the database name must be backtick-quoted, but got: CREATE TABLE my-db.`auto_create_table`(...)
IdentifierQuotingTest.testAlterTableQuotesTableName:125
  expected:<ALTER TABLE [`my-tbl`] add column ...> but was:<ALTER TABLE [my-tbl] add column ...>

Tests run: 8, Failures: 4, Errors: 0
```

The failing-first run was taken with the helper present but the call sites unmodified, so these are genuine assertion failures on unquoted output -- not a compile error, which would have proven nothing about behaviour. After: 8/8.

## Suite

```
sink-connector   311 tests   0 failures   0 errors   2 skipped     (baseline 303/0/0/2)
```

Delta is the new tests only; the five expectation updates are described above.

minguyen9988:
### `disable.drop.truncate` never suppressed a DROP or a TRUNCATE - **Closes #1287**

Pushed as `7d512e592`. This commit **closes #1287**, so please keep it on the merge-time close list.

The setting has existed since the lightweight connector shipped and has never once stopped a destructive statement. There are **two independent root causes**, one per replication path - and a MySQL `TRUNCATE` travels both of them, so fixing either alone still loses the rows.

---

#### Root cause 1 - the DDL path evaluated a flag nothing had written yet

`sink-connector-lightweight/.../cdc/DebeziumChangeEventCapture.java`

```
:761  AtomicBoolean isDropOrTruncate = new AtomicBoolean(false);
:763  if (checkIfDDLNeedsToBeIgnored(DDL, props, sr, isDropOrTruncate))   <- guard READS it here
:769  mySQLDDLParserService.parseSql(DDL, "", clickHouseQuery, isDropOrTruncate);   <- only writer, AFTER
:1433 ... && isDropOrTruncate.get() == true
```

The flag's only writer is `MySQLDDLParserService.java:152`, reached from `parseSql` - which runs at `:769`, after the guard at `:763` has already read it. The guard therefore evaluated a freshly constructed `false` on every DDL event, and the `&& isDropOrTruncate.get() == true` clause could not be satisfied by any input whatsoever.

#### Root cause 2 - the CDC data path had no idea the setting existed

`sink-connector/.../db/batch/PreparedStatementExecutor.java:233` called `metadata.truncateTable(...)` unconditionally.

`grep -rn "DISABLE_DROP_TRUNCATE" sink-connector/src/main` returned **zero hits**: the constant lived only in `SinkConnectorLightWeightConfig.java:32`, with no corresponding entry in `ClickHouseSinkConnectorConfigVariables`. A MySQL `TRUNCATE` also arrives as a CDC record with operation `TRUNCATE`, and that record is executed here - so the table was emptied regardless of the setting.

---

#### The fix

1. `MySQLDDLParserService` gains a static `isDropOrTruncate(String)` that **lexes the raw DDL only** - no parse tree, no listener, no DB access - so a verdict exists *before* translation. Both entry points delegate to one shared `containsDropOrTruncate(List<Token>)`, so the pre- and post-translation classifications cannot drift.
2. `checkIfDDLNeedsToBeIgnored` derives the verdict itself at the top of the method rather than reading a flag some other call is supposed to have set first. **The ordering bug cannot regress, because there is no longer an ordering to get wrong.** The out-parameter is still populated, so it now carries a true value where the caller reads it.
3. `DISABLE_DROP_TRUNCATE` is declared in `ClickHouseSinkConnectorConfigVariables` and defined in `ClickHouseSinkConnectorConfig.newConfigDef()` - which is how *every other setting* reaches `PreparedStatementExecutor`. That method is already handed the `ClickHouseSinkConnectorConfig`, so no new plumbing was invented and no signature changed. `SinkConnectorLightWeightConfig.DISABLE_DROP_TRUNCATE` now references the enum instead of repeating the literal, so one user setting governs both paths.
4. **Both suppression sites log at WARN naming the affected object.** A silently skipped destructive operation is nearly as bad as an unguarded one - the target has just diverged from the source and the operator needs to know which object. The DDL-path log resolves the target from the Debezium record, falling back to the identifiers in the DDL text so a `DROP DATABASE` (which the record does not describe as a table change) is still named.

**The default is not changed.** Only the existing opt-in is made to work.

---

#### Proof - failing first, per root cause

**Root cause 1, BEFORE** (`DebeziumChangeEventCaptureTest`):

```
[ERROR] Tests run: 29, Failures: 3, Errors: 0, Skipped: 0
[ERROR]   DebeziumChangeEventCaptureTest.disableDropTruncateSuppressesDropTable:222 disable.drop.truncate is enabled, so this DROP TABLE must not reach ClickHouse; the guard read an AtomicBoolean whose only writer (parseSql) runs after the guard, so it was always false
[ERROR]   DebeziumChangeEventCaptureTest.disableDropTruncateSuppressesTruncateTable:231 disable.drop.truncate is enabled, so this TRUNCATE must not reach ClickHouse
[ERROR]   DebeziumChangeEventCaptureTest.disableDropTruncateSuppressesDropDatabase:239 a DROP DATABASE is the most destructive statement the setting claims to cover
```

**AFTER:** `[INFO] Tests run: 29, Failures: 0, Errors: 0, Skipped: 0`

**Root cause 2, BEFORE** (`PreparedStatementExecutorDropTruncateTest`):

```
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0
[ERROR]   PreparedStatementExecutorDropTruncateTest.enabledSuppressesCdcTruncate:246 disable.drop.truncate is enabled, so no TRUNCATE may be sent to ClickHouse; PreparedStatementExecutor called metadata.truncateTable() unconditionally and the setting had zero references anywhere in sink-connector/src/main (#1287), statements seen: [TRUNCATE TABLE testdb.orders] expected:<[]> but was:<[TRUNCATE TABLE testdb.orders]>
```

**AFTER:** `[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`

**Controls** - the load-bearing half. With the setting unset or explicitly `false`, `DROP` and `TRUNCATE` still pass through on both paths, and the CDC `TRUNCATE` is still applied at its binlog position between the two flushes rather than at the end of the batch. With the setting enabled, an `ADD COLUMN` and a `CREATE TABLE` still replicate, and the rows either side of a suppressed `TRUNCATE` are still written. Each of these passes both before and after the change - which is what makes them controls rather than more assertions about the fix.

The suppression logging was observed in the passing run:

```
WARN - SUPPRESSED destructive DDL: disable.drop.truncate=true, so this DROP/TRUNCATE was NOT applied
to ClickHouse. Target(s): [contacts]. The ClickHouse object is now out of sync with the source.
DDL: DROP TABLE employees.contacts
```

---

#### Suites

| Suite | Result | Baseline |
|---|---|---|
| `sink-connector`, full | **316 / 0 failures / 0 errors / 2 skipped** | 311/0/0/2 (+5 new tests) |
| `sink-connector-lightweight`, targeted DDL-parser + CDC | **194 / 0 / 0 / 1** | +7 new tests, 0 failures |

No new failures on either module.

minguyen9988:
One more fix. Closes #1288.

## Offset-storage statements interpolated values instead of binding them

Every offset-storage statement except the INSERT built its SQL with `String.format`:

```
DebeziumOffsetStorage.java:78-79    delete from %s where offset_key='%s'
DebeziumOffsetStorage.java:96-99    delete from schema history, by server name
DebeziumOffsetStorage.java:121-122  select max(record_insert_ts) from %s
DebeziumOffsetStorage.java:143-145  select offset_val from %s where offset_key='%s'
DebeziumOffsetStorage.java:256-261  the INSERT -- the only bound path
```

The interpolated key is not a constant: `getOffsetKey` at `:56-60` builds it from the connector's own `name` property with no escaping, so a name carrying a quote character terminates the literal and the remainder is parsed as SQL.

The issue body's code snippets do not match the current file -- they quote a `topic_name` column and a `%s.%s` shape that do not exist here. The line references above are what is actually in the tree.

## Fix

Values are bound as JDBC parameters. `executeSystemQuery` in `DBMetadata` gained a `String... params` tail, so existing two-argument call sites are unchanged.

The table name is an identifier rather than a value, so it cannot be a bind parameter. It is quoted instead, reusing `ClickHouseTableOperationsBase.quoteIdentifier` rather than introducing a second escaping rule -- the lightweight module already depends on that module, and this file already imports `DBMetadata` from it.

The configured table name is **qualified** (`altinity_sink_connector.replica_source_info`), so quoting is applied per dot-separated part to yield `` `db`.`table` ``. Quoting the whole string would produce a single identifier containing a dot, naming a table that does not exist.

Worth recording: `deleteSchemaHistoryTable` at `:96` already wrapped its qualified name in backticks as one string, so it was carrying that second defect independently of the injection one.

## Proof

Failing-first, 5 of 5:

```
deleteOffsetStorageRow_bindsOffsetKey:71 injected DDL reached the driver as SQL:
  delete from altinity_sink_connector.replica_source_info where offset_key='["engine';
  drop table altinity_sink_connector.replica_source_info; --",{"server":"embeddedconnector"}]'
  ==> expected: <false> but was: <true>

deleteSchemaHistoryTable_bindsServerName:114 injected DDL reached the driver as SQL:
  delete from `altinity_sink_connector.replicate_schema_history` where
  JSONExtractRaw(JSONExtractRaw(history_data,'source'), 'server')='engine';
  drop table altinity_sink_connector.replica_source_info; --'

ordinaryConnectorName_producesExpectedStatement:135
  expected: <delete from `altinity_sink_connector`.`replica_source_info` where offset_key=?>
  but was:  <delete from altinity_sink_connector.replica_source_info where offset_key='...'>
```

After: 6 tests, 0 failures. `sink-connector` module 316/0/0/2, unchanged from baseline -- the new tests are in the lightweight module.

One assertion was corrected during the work rather than accommodated. The first version asserted the statement contained no single quote at all, which rejects the constant literals `'source'` and `'server'` inside the `JSONExtractRaw` call the statement legitimately contains -- it would have failed on correct code forever. It now asserts the property that actually matters: the hostile payload appears nowhere in the statement text, and the key is compared against a bind marker.

## Still open from the same area

#1297 (non-atomic offset update) is analysed but not yet fixed. Recording the findings here since they change what the right fix is:

- Transactions are not available. `DebeziumChangeEventCapture.java:280-283` forces `jdbc_ignore_unsupported_values=true` onto the offset-storage JDBC URL precisely so `setAutoCommit(false)` does not throw, with a comment that ClickHouse has no transactions; `SinkConnectorDataSource.java:33` repeats it. A transaction-based fix would be a no-op that reads as correct.
- The naive "stop deleting" reading is also wrong. The offset table is `ReplacingMergeTree(_version) ORDER BY id`, and `id` is a per-insert UUID rather than the offset key, so successive offset rows never collapse -- rows would accumulate and the `FINAL` read would return the whole history.
- The shape that works is insert-then-delete: write the new row first, then delete prior rows for that key excluding the one just written. A crash mid-window leaves two rows, last-wins by `record_insert_ts` and self-healing, instead of zero rows and a lost offset.

minguyen9988:
One more fix. Closes #1297.

## The offset update deleted the stored offset before writing its replacement

Both overloads of `updateDebeziumStorageStatus` in `DebeziumJdbcStorageOperations` issued the pair in this order:

```
:410-411  deleteOffsetStorageRow(offsetKey, props, conn)    <- binlog
          updateDebeziumStorageRow(conn, tableName, ...)
:457-458  deleteOffsetStorageRow(offsetKey, props, conn)    <- LSN
          updateDebeziumStorageRow(conn, tableName, ...)
```

A crash, kill, or connection loss between the two leaves the connector with no stored offset at all. On restart it has nothing to resume from.

## Both obvious fixes are wrong

**Not a transaction** -- there is not one available, and adding one would be a no-op that reads as correct. `DebeziumChangeEventCapture:280-283` forces `jdbc_ignore_unsupported_values=true` onto the offset-storage JDBC URL *precisely so* `setAutoCommit(false)` is silently ignored rather than throwing, with a comment stating ClickHouse has no transactions. `SinkConnectorDataSource:33` repeats it. Wrapping the pair in `setAutoCommit`/`commit` would compile, run, and change nothing.

**Not "stop deleting"** -- the offset table is `ReplacingMergeTree(_version) ORDER BY id` (`docker/config.yml:100-111`), and `id` is a fresh `UUID.randomUUID()` on every insert rather than the offset key, so successive rows for one connector never collapse. The read is `SELECT ... FROM %s FINAL ORDER BY record_insert_ts, record_insert_seq` (`:115`), so dropping the delete would accumulate rows unboundedly and return the whole history rather than the current position.

## Fix

Insert first, then delete what it supersedes, excluding the row just written. The window cannot be closed, so it is pointed the other way: what survives a crash is now a stale offset rather than no offset. A leftover row is harmless -- the read is ordered by `record_insert_ts` so the newer row wins, and the next update clears it.

`updateDebeziumStorageRow` now returns the id it generated so the caller can name the row to spare. The `void` -> `String` change is source-compatible: the only callers are the two sites above plus one test that ignores the return.

`deleteSupersededOffsetRows` is a new method rather than a parameter on the existing `deleteOffsetStorageRow`, because the standalone `deleteOffsets` path (`:216`) genuinely wants to remove every row for the key and must keep doing so.

## Proof

Failing-first, 4 of 4. The main-class change was applied first so the test compiles, with the call sites left in their original order:

```
offsetUpdate_insertsBeforeDeleting:96
  the replacement offset must be written BEFORE the old one is removed,
  otherwise a crash in between loses the offset entirely. Order was:
  [delete from `altinity_sink_connector`.`replica_source_info` ...,
   INSERT INTO `replica_source_info`(id, offset_key, offset_val...]
  ==> expected: <true> but was: <false>

lsnOffsetUpdate_insertsBeforeDeleting:120   same shape on the PostgreSQL path

crashBetweenStatements_leavesAnOffsetRow:149
  Expected java.lang.IllegalStateException to be thrown, but nothing was thrown.

offsetUpdate_stillRemovesSupersededRows:185
  the delete must exclude the row just inserted, was: delete from
  `altinity_sink_connector`.`replica_source_info` where offset_key=?
  ==> expected: <true> but was: <false>
```

The third one is the defect stated exactly: pre-fix the simulated crash never fires, because the delete has already run and the offset is already gone by the time the second statement would execute.

After: 10 tests, 0 failures across both offset-storage test classes -- the 4 new ones plus `DebeziumOffsetStorageSqlInjectionTest` still at 6 of 6, so the #1288 parameterization is intact.

The recording `Connection` harness was extracted out of `DebeziumOffsetStorageSqlInjectionTest` into `OffsetStatementRecorder` and is now shared by both rather than copied -- 124 lines removed from the injection test. It also gained statement-order accessors, which is what this defect needed and the injection test did not.

minguyen9988:
One more fix. Closes #1260.

## The column-drop branch emits SQL ClickHouse cannot parse

Two defects in one statement. Both confirmed against the actual parser with `clickhouse-format`, not from memory:

```
ALTER TABLE `employees` delete column `amount` Float64
  Code: 62 ... failed at position 32 ('column'):
  Expected one of: IN PARTITION, WHERE

ALTER TABLE `employees` DROP COLUMN `amount` Float64
  Code: 62 ... failed at position 46 ('Float64'):
  Expected one of: token, Dot, Comma, INTO OUTFILE, FORMAT, SETTINGS
```

`ClickHouseDbConstants:30` defined the keyword as `delete column`. `DELETE` is the lightweight-delete clause and expects `IN PARTITION` or `WHERE` next, so the column name terminated the parse.

Correcting only the keyword is not enough, as the second line shows. `DROP COLUMN` names a column and stops there, but `createAlterTableSyntax` appended the data type carried in the map on **both** branches.

## Fix

The constant becomes `ALTER_TABLE_DROP_COLUMN = "DROP COLUMN"`, and the type is emitted only on the ADD path, which declares one. The ADD/REMOVE test moved out of the per-entry loop into a local while doing so, since it does not vary across entries.

## Scope -- latent, not live

`createAlterTableSyntax` is only ever invoked with `ADD` (`ClickHouseAlterTable:137`), so **no deployment issues these statements today**. This is a latent defect on a path reachable through the public method.

Worth naming why it survived: the existing test asserted the unparseable string as its expected output.

```java
// ClickHouseAlterTableTest.java:24
String expectedDeleteColumnQuery = "ALTER TABLE `employees` delete column
  `amount` Float64,delete column `occupation` String,...";
```

A test that pins broken output does not detect the defect, it protects it. That assertion is replaced rather than adjusted.

## Proof

Failing-first, 3 of 3 new drop assertions:

```
dropColumn_usesDropColumnKeyword:56
  'delete column' is not ClickHouse syntax -- DELETE expects IN PARTITION or
  WHERE next, so the column name is a SYNTAX_ERROR (Code: 62). Query was:
  ALTER TABLE `employees` delete column `amount` Float64,delete column
  `occupation` String

dropColumn_omitsTheDataType:73
  expected:<...R TABLE `employees` [DROP COLUMN `amount`,DROP COLUMN
  `occupation`]> but was:<...R TABLE `employees` [delete column `amount`
  Float64,delete column `occupation` String]>

dropColumn_stillQuotesIdentifiers:113
  expected:<...LTER TABLE `my-tbl` [DROP COLUMN `my-col`]>
  but was:<...LTER TABLE `my-tbl` [delete column `my-col` Int32]>
```

The two controls passed before the change and after it: the ADD path still declares the data type, and both paths still quote identifiers per the #1264 fix earlier in this branch.

After: 8 tests, 0 failures. Both corrected statements were then round-tripped through `clickhouse-format` to confirm they actually parse, rather than only asserting the string this code happens to produce:

```
ALTER TABLE `employees` DROP COLUMN `amount`,DROP COLUMN `occupation`
ALTER TABLE `my-tbl` DROP COLUMN `my-col`
```

Full `sink-connector` suite: 319 tests, 0 failures, 0 errors, 2 skipped -- the prior 316 plus these 3.

minguyen9988:
One more fix. Closes #1253.

## The shared connection-pool cache is not thread-safe

`HikariDbSource` holds the process-wide pool cache in static collections, and the class contains **zero** occurrences of `synchronized`:

```
:38   private static Map<String, HikariDataSource> instance = new HashMap<>();
:48   private static Map<String, String> currentKey = new HashMap<>();
:185  private static final Set<String> liveEndpoints = new HashSet<>();
:191  private static final Set<String> deadEndpoints = new HashSet<>();
```

Three distinct threads reach them:

- **the batch pool threads** -- `DebeziumChangeEventCapture:1532-1554` creates one `ClickHouseBatchRunnable` per thread on a `ClickHouseBatchExecutor` (a `ScheduledThreadPoolExecutor`) sized by `thread.pool.size`, **default 10** (`ClickHouseSinkConnectorConfig:60`). Each builds a `DbWriter`, reaching `HikariDbSource` via `BaseDbWriter:259` and `:377`.
- **the Debezium event thread** -- `DebeziumChangeEventCapture:889`.
- **the DDL path** -- `DBMetadata`, 12 call sites.

Concurrent `put` on a `HashMap` can corrupt it: resize is not atomic, so two threads resizing together can drop entries, and on older JDKs leave a cycle in a bucket chain that makes a later `get` spin forever.

`getInstance` also did a get-then-put. Two threads missing the same key both built a pool and only the last `put` survived -- the loser's pool was never cached, so it was never closed and held its connections until the process exited. That is now one `compute()` call.

## Also removed: two dead members

- `connectionPool` (`:53`) was declared and **never referenced anywhere** -- grep for the name returns only its own declaration line.
- the instance field `databaseName` (`:72`) was never written. The class is never instantiated: there is no `new HikariDbSource` anywhere in the tree, and the private constructor's body is a single commented-out line.

Dead state sitting next to live shared state is what makes a concurrency review of this file harder than it needs to be.

## Not in scope

The pool-key defect -- keying by database name alone, so two servers addressed under one name evicted each other's pool -- **is already fixed on develop**. This commit does not touch it and does not claim it.

## Proof

Failing-first, 2 of 2:

```
poolCaches_areConcurrentCollections:63->assertConcurrentField:91
  instance is shared by the batch pool threads, the Debezium event thread
  and the DDL path, but is a java.util.HashMap. Concurrent mutation of a
  plain java.util collection can corrupt it during resize.
  ==> expected: <true> but was: <false>

deadMembers_areRemoved:177
  connectionPool was never read or written; it should not survive as
  apparent shared state ==> expected: <true> but was: <false>
```

After: 2 tests, 0 failures. Full `sink-connector` suite: **321 tests, 0 failures, 0 errors, 2 skipped** -- the prior 319 plus these 2.

### A note on why these assertions are structural

A third test was written first: it raced eight threads through the cache and asserted they all saw one pool. **It passed against the unfixed `HashMap`** -- losing that race is timing-dependent, and it did not lose it on that run.

A test that can pass on the broken code proves nothing about the fix, so it was deleted rather than kept for the appearance of coverage. The collection type is the invariant that actually holds every run.

PR #1426, opened 2026-08-31
urgent fix: a keyless source table must not prevent the connector from starting

minguyen9988:
## Why

[#1401](https://github.com/Altinity/clickhouse-sink-connector/pull/1401) added a startup check that **refuses** to replicate a MySQL source holding a table with no PRIMARY KEY and no non-null UNIQUE key. The diagnosis behind it stands — such a table has no row identity in the binlog, so its ClickHouse copy silently diverges — but refusing is the wrong remedy.

**It is not the connector's decision.** A keyless table is often migration bookkeeping nobody reads, or one whose key is already scheduled for a maintenance window. Whether an imperfect copy is acceptable in the meantime is a judgement about the data, which the operator has and the connector does not.

**The blast radius is wrong.** The refusal is per *source*, not per table: one keyless table stops the entire pipeline, including every correctly-keyed table on the same server. On a source with dozens of replicated schemas, a single `alembic_version` row — three columns, written once per migration — takes everything else down with it. The cost lands almost entirely on data that was replicating perfectly well.

## What changes

The check now **reports instead of blocking**. Everything that made the refusal actionable is kept:

- the banner still names every offending table and carries the exact `ALTER TABLE` that fixes it
- it still says `NOT retroactive`, so nobody expects `sql_generate_invisible_primary_key` to repair existing tables
- it is still emitted at ERROR — at startup, at table creation, and periodically during replication

Only the throw is gone. The wording now states plainly that replication is continuing — **"REPLICATING ANYWAY — THESE TABLES WILL DIVERGE"** — so nobody hunts for a startup failure that never happens.

Both escape hatches keep working and now mean what their names say: `table.exclude.list` leaves a table out of replication entirely, and `keyless.table.check.skip=true` silences the reporting.

## Changes

- `UnsupportedSourceException` deleted; both throw sites `log.error` and return.
- `refusalPreExisting` / `refusalTooOld` → `warningPreExisting` / `warningTooOld`, text rewritten from `REFUSING TO REPLICATE`.
- The rethrow in the catch chain removed, so no path escapes `check()`.
- Class javadoc, the `DebeziumChangeEventCapture` call-site comment, the `MySqlDDLParserListenerImpl` note and two parser-test comments corrected — they described a refusal that no longer exists.
- The IT config keeps `keyless.table.check.skip=true`, but its comment no longer claims the ITs *need* it to run. They do not, now that startup is never blocked; it only suppresses a banner in a fixture that is keyless on purpose.

## Tests

**30** in `KeylessTablePreflightTest` (was 28), **160** across the affected suites, all green:

```
KeylessTablePreflightTest        30   0 failures
MySqlDDLParserListenerImplTest  108   0 failures
CreateTableUniqueKeySortKeyTest   9   0 failures
CreateTableNoKeySortKeyTest       7   0 failures
DdlReplayIdempotencyTest          5   0 failures
DataTypeConverterTest             1   0 failures
```

Two of the new tests are **structural rather than behavioural, and deliberately so.** The behavioural test drives every entry path a unit test can reach — the skip override, a non-MySQL source, an unreachable host, an incomplete config — and asserts none throws. But the two sites that actually decided to refuse sit *after* a successful connection and a completed table scan, which no unit test reaches without a live MySQL.

That gap was measured, not assumed: **with the refusal reinstated, the behavioural test still passed and only the structural check failed.** So `check()` is additionally scanned in source for any `throw` statement, and the class is reflected over to prove no refusal-era API survives.

**Failing-first proof** — with the throws restored:

```
testCheckContainsNoThrowStatement  FAILED
  check() must never prevent startup, but line 179 throws:
  throw new UnsupportedSourceException(warningTooOld(version, keyless));
testNoRefusalApiRemains            FAILED
  the refusal exception type must be gone, found
  KeylessTablePreflight$UnsupportedSourceException
```

With the fix in place, both pass.

## Blast radius

Strictly a relaxation: a source that started before continues to start, and a source that was refused now starts and replicates with a loud warning. No schema, DDL translation, or write path is touched — the diff is confined to the preflight, its tests, comments, and one test-resource comment.

Raised by OmniWatcher on behalf of the Jump Trading DBA team, and marked **WIP / DO NOT MERGE**: we do not own this repository, so the merge decision is entirely the maintainers'. Reviewed internally (3/3 approve) before submission.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DebeziumChangeEventCapture.java
@@ -509,10 +509,11 @@ public class DebeziumChangeEventCapture {
         }
         // A source table with no PRIMARY KEY and no non-null UNIQUE key has no
         // row identity in the binlog, so nothing downstream can keep its
-        // ClickHouse copy correct. Enable MySQL's generated invisible primary
-        // key so tables created from now on get one, and refuse to start on a
-        // source that already holds such a table rather than replicating it
-        // wrongly.
+        // ClickHouse copy correct. Name every such table here, with the ALTER
+        // that fixes it, and carry on replicating: whether to accept that
+        // divergence until a key is added is the operator's call, and refusing
+        // would take every correctly-keyed table on the same source down with
+        // it. This call never throws.
         KeylessTablePreflight.check(props);
 
         ClickHouseSinkConnectorConfig config = new ClickHouseSinkConnectorConfig(PropertiesHelper.toMap(props));
--- a/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/KeylessTablePreflight.java
+++ b/sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/KeylessTablePreflight.java
@@ -13,8 +13,8 @@ import java.util.List;
 import java.util.Properties;
 
 /**
- * Refuses to start when the source holds a table whose rows cannot be
- * identified, reading the source and never writing to it.
+ * Reports, loudly, every source table whose rows cannot be identified --
+ * reading the source, never writing to it, and never blocking startup.
  *
  * <p>A table with no PRIMARY KEY and no non-null UNIQUE key has no row
  * identity in the logical schema. InnoDB does give such a table an internal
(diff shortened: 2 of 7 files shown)


minguyen9988:
Request for the maintainers: could this land in the next release cut (2.10.3)?

The commit here (`d4829d38`) is parented directly on `689fd324`, which is exactly the 2.10.2 tag, so `d4829d38` **is** 2.10.2 plus this fix — no rebase or conflict resolution needed to cut a release from it.

We are deliberately **not** asking for the 2.10.2 tag to be moved. That release already has published binaries attached (`clickhouse-sink-connector-client-2.10.2-amd64`/`-arm64`, `clickhouse-sink-connector-lt-2.10.2.jar.zip`, source archives) and `altinityinfra/clickhouse-sink-connector:2.10.2-test-lt` was built from that commit. Retagging would leave those artifacts pointing at source they were not built from, and anyone who already pulled 2.10.2 would silently be running different code under the same version string.

Happy to open a release-branch PR or adjust anything here if that helps get it into the next cut.

PR #1429, opened 2026-09-01
db_compare: MySQL/ClickHouse checksum column-set asymmetry (approach under revision)

minguyen9988:
## Summary

`get_table_checksum_query()` in `clickhouse_table_checksum.py` can silently re-add a column the user explicitly excluded, making the ClickHouse checksum hash a different column set than the MySQL side. When that happens the checksum for the affected table can **never** match, and no `--exclude_columns` / `ignored_columns` value can suppress it.

## Root cause

The un-exclusion rule was introduced for the connector's `is_deleted` / `_is_deleted` pair:

```python
prefixed_column = "_"+row[0]
if row[0] in excluded_columns and prefixed_column in columns_metadata_map:
    logging.info(f"Not excluding column {row[0]} as {prefixed_column} is also excluded")
elif row[0] in excluded_columns:
    logging.info(f"Excluding column {row[0]}")
    continue
filtered_columns_metadata.append(row)
```

The condition matches on the underscore prefix **alone**, so it fires for any excluded column whose `_`-prefixed name happens to exist on the table — not just connector metadata.

A table with a business column `version` and the connector's `_version` hits this: `version` is un-excluded on the ClickHouse side, while `mysql_table_checksum.py` still honours the exclusion. The two sides hash different column sets.

The asymmetry is invisible in the logs — the ClickHouse side logs `Not excluding column version as _version is also excluded` as an INFO line, while the top-level job reports only a checksum mismatch, which reads as a data divergence.

## Reproduction

A table with both `version` and `_version`, excluding `version`:

```
Column set hashed   Result
-----------------   ------
MySQL side          id, bunit, date                (version excluded, as requested)
ClickHouse side     id, bunit, date, version       (version silently re-added)
```

Observed on a table of 1,407,663 rows where every column value is byte-identical between MySQL and ClickHouse. Re-computing the tool's own hash arithmetic over both sides confirms the data agrees and only the column set differs:

```
Column set                          MySQL bucket hash    ClickHouse bucket hash   Match
--------------------------------    -----------------    ----------------------   -----
id,bunit,date        (both sides)      429610042465157         429610042465157     yes
id,bunit,date,version (both sides)     429390065739979         429390065739979     yes
```

Both sides agree under *either* consistent column set. The mismatch exists only in the asymmetric set the tool actually builds.

Natural control on the same schema: a sibling table with `_version` but **no** `version` column passes cleanly.

## Fix

Restrict the un-exclusion rule to columns the connector itself manages:

```python
SINK_METADATA_COLUMNS = frozenset(
    {"_sign", "_version", "is_deleted", "_is_deleted", "__is_deleted"}
)
```

```python
if (row[0] in excluded_columns
        and prefixed_column in columns_metadata_map
        and row[0] in SINK_METADATA_COLUMNS):
```

- `is_deleted` remains in the set, so the original `is_deleted`/`_is_deleted` behaviour the rule was written for is unchanged.
- `version` is not connector metadata, so a user-supplied exclusion is now honoured.
- Fails safe: an unrecognised column name is now excluded as asked, rather than silently re-added.

The set matches the metadata columns the tooling already treats as connector-owned — `clickhouse_table_checksum.py` defaults `--exclude_columns` to `_sign,_version,is_deleted,_is_deleted`, and `top_level_table_checksum.py` always passes `_version,is_deleted,_is_deleted,__is_deleted`. A user column such as `version` never enters the excluded list except via explicit user configuration, which is precisely the intent that must not be overridden.

## Verification

Simulating the filter over the real affected column list (`id, risk, bunit, date, _version, is_deleted, version, clientupdatetime, action, state, user`) excluding `_version, is_deleted, _is_deleted, __is_deleted, version, clientupdatetime, action, state, user, risk`:

```
             Columns hashed on the ClickHouse side
before fix   ['id', 'bunit', 'date', 'version']     <- version wrongly re-added
after fix    ['id', 'bunit', 'date']                <- matches the MySQL side
```

Regression check on the pair the rule exists for (`id, val, is_deleted, _is_deleted`):

```
before fix   ['id', 'val', 'is_deleted']
after fix    ['id', 'val', 'is_deleted']            <- unchanged
```

`python -m py_compile` passes on the modified file.

## Scope

One file, two hunks, no behaviour change for any table without a business column colliding with a connector metadata name. Tables affected by the bug today are only those carrying a column whose `_`-prefixed twin also exists.

## Suggested follow-up

A regression test covering a table with both `version` and `_version`, asserting `version` is absent from the generated select. This is the case that reads as correct under any single-column test, which is why it survived.

---

Filed by OmniWatcher, an AI agent, on behalf of the Jump Trading DBA team, at the request of a senior DBA. Marked **WIP / DO NOT MERGE** because we do not own this repository — please treat it as a proposal for maintainer review rather than a ready-to-merge change, and tell us if you would prefer a different shape for the fix (for example, deriving the metadata set from existing constants). Happy to add the regression test in this PR if you want it here rather than as a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/release-notes/2.10.1.md
+++ b/release-notes/2.10.1.md
@@ -1,0 +1,6 @@
+## What's Changed
+* fix: history mode loses ~95% of rows when a CREATE is replayed by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1410
+* fix: history-mode DELETE lost -- _operation dropped from INSERT, and the delete races its own batch by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1411
+
+
+**Full Changelog**: https://github.com/Altinity/clickhouse-sink-connector/compare/2.10.0...2.10.1
--- a/release-notes/2.10.2.md
+++ b/release-notes/2.10.2.md
@@ -1,0 +1,12 @@
+## What's Changed
+* fix: /metrics served with no Content-Type, breaking Prometheus 3.x scrapes by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1413
+* fix: INT8 (Byte) hard-cast to Integer fails every TINYINT batch by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1414
+* fix: ARRAY binding hard-casts to ArrayList, breaking every other List type by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1415
+* fix: initial snapshot never finishes -- NPE on every unparseable record by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1416
+* fix: MySQL DOUBLE auto-created as Float32, silently truncating precision by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1417
+* fix: support PostgreSQL infinity values in timestamptz by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1418
+* fix: batch containing a missing ClickHouse table is silently dropped and the offset advances by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1419
+* fix: _version stores UInt64 max when the source has no GTID, permanently winning RMT dedup by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1420
+* Added release notes for 2.10.1 release by @subkanthi in https://github.com/Altinity/clickhouse-sink-connector/pull/1412
+
+**Full Changelog**: https://github.com/Altinity/clickhouse-sink-connector/compare/2.10.1...2.10.2
--- a/release-notes/2.10.3.md
+++ b/release-notes/2.10.3.md
@@ -1,0 +1,5 @@
+## What's Changed
+* urgent fix: a keyless source table must not prevent the connector from starting by @minguyen9988 in https://github.com/Altinity/clickhouse-sink-connector/pull/1426
+
+
+**Full Changelog**: https://github.com/Altinity/clickhouse-sink-connector/compare/2.10.2...2.10.3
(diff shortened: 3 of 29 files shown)


minguyen9988:
## Change of direction — please do not review the current diff

A senior DBA on our side pointed out that my framing was wrong, and he is right. Summarising here rather than silently force-pushing, so the history stays honest.

**What I got wrong.** I treated the un-exclusion rule as a pure bug and proposed making ClickHouse honour the exclusion. But `version` in the affected table is a **real MySQL generated column** carrying business data:

```
`version` bigint GENERATED ALWAYS AS (json_unquote(json_extract(`risk`,'$.Version'))) STORED
```

So the ClickHouse behaviour — keeping the real business column and dropping only the connector's shadow `_version` — is arguably the **desirable** one. My patch would have made the tool silently stop checksumming a real MySQL column, which is worse for data integrity than the symptom it fixes.

**What the actual defect is.** Not the rule itself, but that only one side applies it:

- `clickhouse_table_checksum.py` (~line 116): keeps `X` when `_X` also exists.
- `mysql_table_checksum.py` (lines 75-77): no such rule, plain `continue`.

The two sides therefore hash different column sets whenever a table has both `X` and `_X`. The fix should restore symmetry by **inclusion** on the MySQL side, not by exclusion on the ClickHouse side.

Note the two sides are also fed different exclude lists by `top_level_table_checksum.py`: the MySQL command (~line 148) receives only the user-supplied ignores, while the ClickHouse command (~line 163) additionally receives `_version,is_deleted,_is_deleted,__is_deleted`. A symmetric rule needs the MySQL side to know about the connector metadata columns too.

One edge case worth flagging before anyone writes it: a naive symmetric rule keyed on *both `X` and `_X` appearing in the exclusion list* would also start keeping `is_deleted` on tables that have `is_deleted` but no `_is_deleted` column — a behaviour change against today. Whatever shape you prefer should decide that case deliberately.

**Our immediate situation is resolved without any tool change.** The column was only in our ignore list because it was added while iterating to get a job green; the data was never divergent. Removing that one ignore line makes both sides hash `id, bunit, date, version`, which we verified is byte-identical across all four systems (2 MySQL sources, 2 ClickHouse replicas, 1,407,663 rows, hash `3023387072080231`). So there is no time pressure from us.

**How would you like to proceed?** Happy to either (a) repoint this PR at the MySQL-side symmetric fix, or (b) close it and leave the report as an issue if you would rather own the fix shape yourselves. The asymmetry is real either way and silently produces an unfixable checksum mismatch, so it seemed worth reporting even though we no longer need it patched.

Apologies for the churn.

PR #1428, opened 2026-09-01
fix: post-snapshot heartbeat offset is never committed, so the snapshot never finishes (#1379)

minguyen9988:
Reopens the underlying defect behind #1379. The reporter confirmed on 2026-08-31 that the symptom persists on **2.10.2**, which already contains the fix from #1416.

## This is not the defect #1416 fixed

#1416 stopped a `NullPointerException` on records the parser cannot convert and turned the crash-spam into a clean, logged skip. That was correct as far as it went — but the record is **still dropped**, and dropping it is the actual bug. The reporter's 2.10.2 log shows #1416's new WARN firing exactly as designed, with the snapshot still never finishing.

## Root cause

A Debezium batch carries more than row changes. A heartbeat has a `Struct` value and no `op` field, so `SourceRecordParserService.parse` returns null for it — that null is contract, not failure. Such a record is then dropped outright: never `markProcessed`, never `markBatchFinished`. **Only records that become a row are ever acknowledged.**

That silently strands the end-of-snapshot state. Debezium marks the snapshot complete *after* the last snapshot row is emitted — `RelationalSnapshotChangeEventSource#createDataEvents` calls `preSnapshotCompletion`/`postSnapshotCompletion` once the row loop is done — so every snapshot **row** still carries `snapshot=INITIAL, snapshot_completed=false`. The completed state rides only on records emitted *after* the snapshot, and on an idle source those are exclusively heartbeats.

So `replica_source_info` stays pinned at `snapshot=INITIAL, snapshot_completed=false` forever, which is both halves of the report:

- the status view never shows the snapshot finishing, even though the log says `Snapshot completed` and every row is in ClickHouse;
- on restart `InitialSnapshotter#shouldSnapshotData` reads `snapshotInProgress=true` and re-runs the entire snapshot — the reporter's "restarting the container retriggers the whole process from beginning".

The reporter's own log shows the mechanism directly. The dropped heartbeat carries:

```
sourceOffset={lsn=74130877960, txId=3322, ts_usec=1788182207848583}
```

with **no snapshot keys at all** — `PostgresOffsetContext#getOffset` omits `snapshot`, `last_snapshot_record` and `snapshot_completed` once `postSnapshotCompletion` has cleared the snapshot type — while the stored offset still has `snapshot=INITIAL`. Acknowledging that one heartbeat is precisely what lets a restart skip the snapshot.

## The fix

Acknowledge the offset of a record that produced no row — but **only when nothing is in flight**.

[...]

PR #1430, opened 2026-09-02
fix: the offset table name arity guard rejects every valid value (#1379)

minguyen9988:
Follow-up to #1428, on top of that branch. Fixes the last loose end in #1379: the WARN the reporter kept asking about, plus a k8s configmap that cannot start the connector at all.

`WIP: DO NOT MERGE` because this targets `fix-1379-control-record-offset`, which is itself still open as #1428 — merge that first.

## 1. `getOffsetStorageDatabaseName` rejects every valid value

The reporter kept seeing this and went looking for a schema-history misconfiguration:

```
WARN [DbWriter] - Skipping creating offset schema history table as the
query was not provided in configuration
```

There is nothing to find. On PostgreSQL there is **no schema history at all** — `PostgresConnectorConfig` extends `RelationalDatabaseConnectorConfig`, not the Historized variant, because Postgres rebuilds schema from logical decoding and the live catalog on every start. `config_postgres.yml` correspondingly sets zero `schema.history.*` keys, correctly.

The WARN is about neither schema history nor a missing query. It is an arity check on the offset **table name**, and it is off by one:

```java
String[] offsetStorageDatabaseNameArray = offsetSchemaHistoryTable.split("\\.");
if (offsetStorageDatabaseNameArray.length <= 2) {   // <-- rejects the correct value
    log.warn("Skipping creating offset schema history table as the "
            + "query was not provided in configuration");
    return null;
}
String offsetStorageDatabaseName = offsetStorageDatabaseNameArray[0];
String offsetStorageTableName = offsetStorageDatabaseNameArray[1];
```

The property is a qualified `database.table` pair, so a correct value has exactly **two** parts. Measured on the value every sample config in this repo ships:

```
"altinity_sink_connector.replica_source_info".split("\\.")
    -> length 2
    -> guard "<= 2"  rejects  (returns null)
    -> guard "<  2"  accepts  (db=altinity_sink_connector,
                               table=replica_source_info)
```

So the method returned `null` for every correctly configured deployment — and note the two lines immediately after the guard read `[0]` and `[1]`, i.e. they assume the two-part shape the guard just refused. They were only ever reachable for a three-part name that no config in the tree uses.

**Why this is not an outage, and what the damage actually is.** On the lightweight path `DebeziumChangeEventCapture` calls `DebeziumJdbcStorageOperations.createDatabaseForDebeziumStorage` at startup, which creates the same database from the same property with a correct two-part parse. The `DbWriter` path is dead code that has been silently doing nothing. The real cost is the message: emitted on every start, naming a mechanism it is not about, and asserting a query was not provided when one was. It cost this issue a round trip.

Fixed the comparison and reworded both warnings to name the actual property and the actual condition.

## 2. The parse is now testable without a database

Extracted to a package-private static. `DbWriter`'s constructor opens a connection, so a test that built a writer just to read a config string spent the full connection-retry budget first — the first version of this test took minutes and drowned in `Retry Number: 9 of 10 Error creating Database`. The new one runs in 0.3s with no Docker.

## 3. Second defect: a k8s configmap that cannot start the connector

`k8s/clickhouse-sink-connector-lt-cm1-configmap.yaml` sets **only** the deprecated alias `offset.storage.jdbc.offset.table.name`.

Debezium accepts that alias. This connector does not read it: `DebeziumJdbcStorageOperations` composes its lookup key by literal concatenation of `Field.name()`, which yields only the canonical name, and throws `IllegalArgumentException` when the lookup comes back null. Probed against the pinned Debezium 3.1.3.Final:

```
OFFSET_STORAGE_PREFIX    = [offset.storage.]
PROP_TABLE_NAME.name()   = [jdbc.table.name]
composed lookup key      = [offset.storage.jdbc.table.name]   <- canonical only
```

Switched to the canonical key. Both YAML layers — the configmap and the `config.yml` string it embeds — were re-parsed after the edit. The alias form for `.ddl`/`.delete`/`.select` is left alone: those are read by Debezium itself, which honours the aliases.

The two other configs that mention the alias for `table.name` (`docker/config_oracle.yml` and the systemd deploy template) already set the canonical key alongside it, so they work. Not touched.

## Verification

- `OffsetStorageDatabaseNameTest` — 5 cases, no Docker, 0.3s. The shipped two-part default resolves, a three-part name still resolves to its first segment, and an unqualified / unset / empty value is still rejected.
- **Failing-first verified.** Restoring only the old comparison fails `testShippedDefaultResolves` with `expected: <altinity_sink_connector> but was: <null>`, while the other four still pass — so the fix cannot be satisfied by simply accepting everything.
- `sink-connector` unit suite: 262 run, 0 failures, 19 errors. All 19 are testcontainers `Could not find a valid Docker environment` on the build host; none are from this change.

Docker-dependent integration suites were not run locally; CI exercises them.

## Still open on #1379, not in this PR

The restart-survival integration test. `PostgresInitialDockerIT` asserts `offsetValue.contains("snapshot")`, which passes on both `"snapshot":"INITIAL"` and any completed shape, and it never restarts the engine — so the re-snapshot symptom is outside its coverage entirely. #1428's two tests are unit-level. A `PostgresSnapshotResumeIT` that stands up Postgres + ClickHouse, waits past one heartbeat interval, asserts exactly one offset row and `snapshot_completed`, restarts, and asserts the snapshot does not re-run would close it. Worth doing separately rather than widening this PR.

Created by OmniWatcher on behalf of the Jump Trading DBA team; requires human review before merging.


minguyen9988:
sorry I merged wrongly

PR #1432, opened 2026-09-03
fix: eliminate the DDL/DML race that silently drops column values

minguyen9988:
fix: NULL-valued columns dropped from the INSERT, so ClickHouse writes the DEFAULT (#1389 follow-up)

A column that is genuinely NULL in the source row is omitted from the
generated INSERT entirely, so ClickHouse applies the column DEFAULT
(0 / '' / 1970-01-01) instead of NULL. Row counts still match, so a
count-based checksum reports the table clean; only a value-level
comparison catches it.

Root cause
----------
ClickHouseStruct#setAfterStruct builds the "modified fields" list by
keeping only the fields whose value is != null:

    for (Field f : schemaFields) {
        if (s.get(f) != null) {          // <-- NULL columns filtered out
            this.afterModifiedFields.add(f);
        }
    }

PR #1389 then made QueryFormatter#createColumns decide INSERT membership
from that list, to stop a record buffered before an ALTER TABLE ADD
COLUMN from NULL-filling the newly added column. That fix is correct for
the case it targeted, but the list it consults conflates two different
things:

  - a column the record does not carry at all (pre-ALTER)  -> must be
    omitted, so ClickHouse applies the DEFAULT the ALTER established;
  - a column the record carries whose value is NULL        -> must be
    bound, so the cell lands as NULL.

Both look identical in the value-filtered list, so the second case
inherited the first case's treatment and its value stopped being written.

The two are distinguishable: a pre-ALTER record does not carry the column
in its SCHEMA, whereas a NULL-valued column is in the schema and only
absent from the value-filtered list. The schema is therefore the correct
authority for INSERT membership; the value-filtered list never was.

Fix
---
getInsertQueryUsingInputFunction takes the record's unfiltered schema
fields as a new trailing argument and decides membership from them,
falling back to the previous behaviour when null. The existing 7- and
8-argument signatures are untouched, so no caller is forced to change.
GroupInsertQueryWithBatchRecords resolves the schema from whichever image
the value-filtered list came from, by identity, so an UPDATE's
before-image is never resolved against the after-image's schema.

The binder needed no change: PreparedStatementFieldMapper already calls
ps.setNull() when struct.get(colName) returns null. The value never
reached it because the column had no placeholder to bind.

Impact observed in production
-----------------------------
On txnrepo-sink-uat after the 2.10.1 -> 2.10.3 upgrade, with the
connector logging "Column index missing for column" and writing the row
anyway:

    59,929  txnrepo_uat.event                kafka_offset
     3,264  aerion_uat.aerion_break_detail   break_original_date, estimate_resolution_date
     1,951  aerion_uat.aerion_rec            jump_count, clearing_count, rec_end_datetime, fail_reason
       433  trade_uat.enriched_trade         capped_by

Verification (maven 3.9.11 / JDK 17)
------------------------------------
  QueryFormatterNullValueDropTest      4/4 pass   (formatter contract)
  NullValueColumnDropTest              4/4 pass   (production entry point)
    with the fix neutralised           5 of 8 FAIL, emitting
                                       INSERT INTO `event`(`id`,`request`,`_version`,`is_deleted`)
                                       -- kafka_offset absent from the statement
  sink-connector module suite          271 run, 0 failures, 20 errors
  untouched baseline 3c0759b1          263 run, 0 failures, 20 errors

The delta between the two runs is exactly the 8 tests added here. The 20
errors are pre-existing and environmental: every one is testcontainers
"Could not find a valid Docker environment" on a host with no Docker
daemon, and they reproduce identically on the untouched baseline.

Both new suites include a regression guard asserting that a record whose
schema genuinely lacks the column is still omitted, so #1389 cannot be
undone by this change.

Co-authored-by: OmniWatcher <dba@jumptrading.com>


minguyen9988:
## Correction from the author — the end-to-end reproduction does NOT confirm this diagnosis

I am flagging this against my own PR before anyone spends review time on it.

I wrote an end-to-end integration test (`NullColumnValueRoundTripIT`, MySQL 8.0.36 -> connector -> ClickHouse 24.8, real containers) covering the four production failure shapes: NULL on insert, NULL blob, UPDATE-to-NULL, and several NULL columns in one row.

**It passes with the fix in place — and it also passes with the fix reverted.** That is a negative result on my own hypothesis, not a confirmation:

```
NullColumnValueRoundTripIT, fix applied     4/4 pass
NullColumnValueRoundTripIT, fix neutralised 4/4 pass   <-- should have failed
```

The generated statement in the IT already binds every column before the fix:

```
INSERT INTO `event`(`id`,`kafka_offset`,`request`,`recorded_at`,`_version`,`is_deleted`) VALUES (?,?,?,?,?,?)
```

So on this path a NULL-valued column is **not** dropped, and the unit-level failure I based the diagnosis on does not reproduce end to end.

### What this means

The unit tests in this PR are genuine — `QueryFormatterNullValueDropTest` and `NullValueColumnDropTest` both fail without the change and pass with it, and the statement they produce really does omit `kafka_offset`. But they reach `createColumns` with a value-filtered field list, and the live connector evidently does not, at least for a plain MySQL `INSERT ... VALUES (null, ...)`. Debezium appears to emit the full after-image with explicit nulls, so `setAfterStruct`'s `s.get(f) != null` filter does not strip the column the way my unit fixtures assume.

**Therefore: the code change here is defensible as hardening (schema is the right authority for INSERT membership, not a value-filtered list), but it is NOT proven to be the cause of the production data loss, and it should not be described as the fix for it until something reproduces that loss.**

### What is solid, independent of the above

- The production divergence is real and value-level verified against the source MySQL host: `txnrepo_uat.event.kafka_offset` 3,348 rows in one day, `trade_uat.enriched_trade.capped_by` 240, plus a fully silent blob loss on `trade_uat.changelog`.
- The connector's own error log records `Column index missing for column` 65,577 times across four tables, then writes the row anyway.

So the symptom and the error are established; the mechanism I proposed is not.

### What I am doing next

Finding the input shape that actually produces the dropped column — the error log proves the connector reaches that branch in production, so something in the real record stream differs from my fixture. Candidates I have not yet eliminated:

1. A stale `columnNameToIndexMap` reused across a schema change, so the map predates a column the current table has.
2. `getWithoutDefault` under `non.default.value`, which yields a different field set than `struct.get`.
3. A path where the before-image's filtered list is used against the after-image's schema.
4. Something specific to the bitemporal/history-mode write path these tables use.

Until one of those is demonstrated, treat this PR as unverified. I would rather say that now than have it merged as a fix for a bug it may not fix.

minguyen9988:
fix: eliminate the DDL/DML race that silently drops column values

The production loss is a race between DDL and DML, not a NULL-handling
bug. A record captured under the pre-ALTER schema is bound against the
post-ALTER table: the row inserts successfully, the affected column takes
the ClickHouse DEFAULT (0 / 1970-01-01 / empty) instead of its real
value, and row counts stay identical on both sides. Nothing fails. A
daily value-level checksum was the only thing that noticed, two days
later.

Five defects, each of which alone reopens the window.

1. pause()/awaitQuiescent() was a check-then-act race. beforeExecute()
tested isPaused and incremented activeBatches as two separate steps, so a
pool thread could read isPaused==false, be overtaken by pause() plus a
drain that saw zero in-flight batches and returned TRUE, and then run its
batch against the just-altered table. Both operations now happen in one
critical section on a private monitor, so a successful drain means no
batch is running AND none can start. pause()/resume()/afterExecute()
publish through the same monitor and use wait/notify rather than
sleep-polling.

2. The drain applied the DDL anyway on timeout. drainBeforeDDL() logged a
warning and proceeded, which is precisely the corrupting interleaving it
exists to prevent. It now throws, routing into the existing DDL retry
path. It also pauses BEFORE draining the queue -- draining while the pool
can still pick up work is unwinnable on a busy table.

3. A failed DDL left the writer paused forever. executor.resume() sat
after performDDLOperation() with no finally; with (2) now able to throw,
that path would stop replication dead. The resume is unconditional.

4. Unresolvable DDL left every cache stale. Invalidation only covered
table names getTableNamesFromDDL() could resolve. Unresolvable or failed
cases now call CacheInvalidationManager.invalidateAll(), which bumps a
global epoch folded into every per-table version -- a per-table map alone
cannot express "everything is stale", since a table never individually
invalidated reads version 0 before and after the sweep.

5. Dropped columns were logged and written anyway. A data column with no
placeholder was logged at ERROR and skipped; a column absent from the
record but present in the cached map was bound to NULL, overwriting the
stored value on an UPDATE (the RENAME/DROP half of #1389). Both now throw
StaleSchemaCacheException, failing the batch into the retry path.
Connector-managed and history-mode columns keep the previous behaviour.

Defence in depth: GroupInsertQueryWithBatchRecords re-reads table
metadata when the record carries a column the cached map does not know
about. QueryFormatter decides INSERT membership from the record schema
rather than the value-filtered modified-fields list.

Verification (maven 3.9.11 / JDK 17, rootless podman):
  PauseDrainAtomicityTest      2/2 pass; with fix reverted 1/2 FAIL
  StaleSchemaCacheIT           4/4 pass, 465s, real MySQL 8.0.36 ->
                               ClickHouse 24.8, DDL against a LIVE
                               connector with a cached writer
  PauseDrainRaceTest           3/3 pass
  sink-connector suite         276 run, 0 failures, 20 errors
  baseline 3c0759b1            263 run, 0 failures, 20 errors

The 20 errors are pre-existing testcontainers no-Docker failures and
reproduce identically on the baseline.

PauseDrainRaceTest passes against the BROKEN executor too (verified by
reverting), so it is a behavioural guard, not the proof.
PauseDrainAtomicityTest is the proof: it asserts atomicity directly by
holding the monitor and does fail on the broken code. The earlier
NullColumnValueRoundTripIT is superseded for the same reason -- it wrote
NULLs to a fresh table, where no staleness exists.

Destructive-Op-Check: sites=1; result=pass
Co-authored-by: OmniWatcher <dba@jumptrading.com>

PR #1437, opened 2026-09-09
2.11.0 

minguyen9988:
# Architecture & End-to-End Flow Specification: ClickHouse Sink Connector (2.11.0)

## Executive Summary

This document provides a comprehensive, ground-truth architectural specification and method-by-method operational mapping of the `clickhouse-sink-connector` on branch `2.11.0`. It covers both operating runtimes:
1. **Embedded Lightweight Mode (`sink-connector-lightweight`)**: In-process Debezium engine streaming MySQL/PostgreSQL binlog/WAL directly to ClickHouse without Kafka brokers.
2. **Kafka Connect Sink Mode (`sink-connector`)**: Standard Kafka Connect `SinkTask` consuming from Kafka partitions and writing batches to ClickHouse.

The purpose of this specification is to establish a rigorous, idea-level mapping of all control paths, data pipelines, state machines, concurrency models, and failure domains. This mapping serves as the foundation for identifying structural incoherencies, race conditions, and design complexity traps that have led to recurring operational defects, and provides the architectural blueprint required for a coherent system redesign.

---

### 1.1 The Prime Directive
The connector is a **replication engine** from transactional source databases (MySQL, PostgreSQL) to an analytical columnar store (ClickHouse).
1. **The source database is the absolute source of truth.** The source data defines what the data is. The connector never assumes the source is in error.
2. **ClickHouse is the replica and must conform to the source.** The replica schema, columns, types, nullability, and row state exist exclusively to mirror the source.
3. **Every divergence must be resolved by correcting ClickHouse to match the source.** Divergences are never resolved by ignoring, modifying, or reinterpreting source events.
4. **The connector is an active enforcer, not merely a passive reporter.** When ClickHouse schema or column definitions prevent storing source data (e.g. `MATERIALIZED` column shadowing source values), the connector automatically alters the ClickHouse schema to enable ingestion.

### 1.2 Architectural Dualism

```
+----------------------------------------------------------------------------------------------------+
|                                    OPERATING TOPOLOGY MODES                                        |
+----------------------------------------------------------------------------------------------------+
|                                                                                                    |
|  [MODE A: Embedded Lightweight Standalone]                   [MODE B: Kafka Connect Plugin]        |
|                                                                                                    |
|   +--------------------------+                                 +--------------------------+        |
|   |   MySQL / PostgreSQL     |                                 |   Kafka Topic Partitions |        |
|   +------------+-------------+                                 +------------+-------------+        |
|                |                                                            |                      |
|                v (Binlog / WAL)                                             v (SinkRecord)         |
|   +--------------------------+                                 +--------------------------+        |
|   | Debezium Embedded Engine |                                 |  ClickHouseSinkTask      |        |
|   +------------+-------------+                                 +------------+-------------+        |
|                |                                                            |                      |
|                v (ChangeEvent)                                              v (put())              |
|   +--------------------------+                                 +--------------------------+        |
|   | DebeziumChangeEvent     |                                 | DeDuplicator (Optional)  |        |
|   | Capture                  |                                 +------------+-------------+        |
|   +------------+-------------+                                              |                      |
|                |                                                            v                      |
|                +-----------------------------+------------------------------+                      |
|                                              |                                                     |
|                                              v (ClickHouseStruct batches)                          |
|                             +--------------------------------+                                     |
|                             | LinkedBlockingQueue (Handoff)  |                                     |
|                             +----------------+---------------+                                     |
|                                              |                                                     |
|                                              v (poll())                                            |
|                             +--------------------------------+                                     |
|                             | ClickHouseBatchExecutor        |                                     |
|                             | (ScheduledThreadPoolExecutor)  |                                     |
|                             +----------------+---------------+                                     |
|                                              |                                                     |
|                                              v (Task Threads)                                      |
|                             +--------------------------------+                                     |
|                             | ClickHouseBatchRunnable        |                                     |
|                             | / ClickHouseBatchWriter        |                                     |
|                             +----------------+---------------+                                     |
|                                              |                                                     |
|                                              v (JDBC PreparedStatement Batch)                      |
|                             +--------------------------------+                                     |
|                             | ClickHouse Cluster / Instance  |                                     |
|                             | (ReplacingMergeTree Engines)   |                                     |
|                             +----------------+---------------+                                     |
+----------------------------------------------------------------------------------------------------+
```

---

## 2. Core Component Directory & Responsibilities

| Component | Module | Responsibilities |
|---|---|---|
| `ClickHouseDebeziumEmbeddedApplication` | `lightweight` | Main entry point for standalone service; loads properties and YAML; initializes Guice dependency injection (`AppInjector`); coordinates lifecycle. |
| `DebeziumChangeEventCapture` | `lightweight` | Embedded Debezium engine harness; manages CDC loop, monotonic sequence numbers, DDL interception, pre-DDL pipeline draining, and control-record offset commits. |
| `ClickHouseSinkTask` | `sink-connector` | Kafka Connect sink task entry point; manages partition assignment, consumes `SinkRecord` batches, invokes deduplication, tracks durable offsets in `preCommit()`. |
| `ClickHouseBatchExecutor` | `sink-connector` | Subclass of `ScheduledThreadPoolExecutor`; introduces thread-safe gate monitor with `pause()`, `resume()`, and `awaitQuiescent()` for DDL barrier synchronization. |
| `ClickHouseBatchRunnable` | `sink-connector` | Worker thread execution loop for multi-threaded mode; drains queues, manages database overrides, builds `DbWriter`, formats queries, and executes JDBC flushes. |
| `ClickHouseBatchWriter` | `sink-connector` | Single-threaded synchronous writer (used when `threadPoolSize == 1`); processes batches inline without thread dispatch. |
| `DbWriter` | `sink-connector` | Encapsulates JDBC connection, ClickHouse metadata cache (column name to type mapping), auto-table creation, and table engine resolution. |
| `GroupInsertQueryWithBatchRecords` | `sink-connector` | Groups incoming records by target query template; performs schema staleness checks (`refreshIfRecordHasUnknownColumn`); enforces column writability; handles UPDATE splitting. |
| `QueryFormatter` | `sink-connector` | Constructs parameterized ClickHouse `INSERT INTO table(cols) VALUES (?,?,...)` queries; builds column-to-index parameter maps. |
| `PreparedStatementFieldMapper` | `sink-connector` | Maps Kafka Connect struct fields to JDBC statement parameter indices; converts data types; populates engine columns (`_version`, `_sign`, `is_deleted`). |
| `PreparedStatementExecutor` | `sink-connector` | Dispatches record batches into JDBC `PreparedStatement`; detects sorting key mutations on UPDATEs and synthesizes delete tombstones; triggers `executeBatch()`. |
| `ClickHouseDataTypeMapper` | `sink-connector` | Central type conversion matrix; maps Debezium/Kafka Connect types to ClickHouse types for DDL generation and statement binding. |
| `CacheInvalidationManager` | `sink-connector` | Global singleton managing monotonic invalidation version counters per table and global invalidation epochs; tracks columns proven absent. |
| `DebeziumOffsetManagement` | `sink-connector` | Tracks in-flight batches; serializes offset updates through Debezium `RecordCommitter` under `OFFSET_COMMIT_LOCK`; checks pipeline quiescence. |
| `PostgresSchemaChangeDetector` | `lightweight` | Detects schema drift in PostgreSQL CDC by comparing Debezium event schemas against ClickHouse `system.columns` using TTL-cached metadata. |
| `PostgresSchemaReconciler` | `lightweight` | Issues automated `ALTER TABLE ... ADD COLUMN` statements on ClickHouse when PostgreSQL schema drift is detected. |
| `DDLParserFactory` / Services | `lightweight` | ANTLR4-based DDL parsers (`MySQLDDLParserService`, `PostgreSQLDDLParserService`) that translate source DDL into ClickHouse DDL. |
| `BinLogHistory` | `sink-connector` | Handles bitemporal SCD Type 2 history logging (`_valid_from`, `_valid_to`, `_operation`, `is_deleted`) in `binlog_history` tables. |

---

### 3.1 Phase 1: Initialization & Engine Startup

1. **Bootstrap & Property Resolution**:
   - `ClickHouseDebeziumEmbeddedApplication.main()` is invoked.
   - Configuration is loaded from `config.properties`, YAML files, or environment variables (`EnvironmentConfigurationService`).
   - `SinkConnectorLightWeightConfig` and `ClickHouseSinkConnectorConfig` validate settings (database overrides, credentials, thread pool size).
2. **Source Preflight Validation**:
   - For MySQL sources, `KeylessTablePreflight.checkKeylessTables()` queries source metadata (`information_schema.tables`, `innodb_indexes`) to identify tables lacking primary keys. Warns that keyless tables will collapse to a single row under ClickHouse `ReplacingMergeTree` unless generated invisible primary keys (GIPK) are enabled.
3. **Target Metadata & Storage Initialization**:
   - `DebeziumJdbcStorageOperations.createSchemaHistoryTable()` creates the schema history table in ClickHouse.
   - `DebeziumJdbcStorageOperations.createViewForShowReplicaStatus()` sets up the replica status monitoring view.
   - If replication history mode is enabled (`replication.history.enable=true`), `ClickHouseAutoCreateTable.createHistoryDatabase()` and `createHistoryTable()` ensure the `binlog_history` database and table exist.
4. **Execution Harness Setup**:
   - A `LinkedBlockingQueue[List[ClickHouseStruct]] records` is instantiated with capacity `max.queue.size` (default unbounded).
   - If `threadPoolSize exceeds 1`, a `ClickHouseBatchExecutor` is initialized with scheduled worker threads running `ClickHouseBatchRunnable` at a fixed rate (`buffer.flush.time`).
   - If `threadPoolSize == 1`, `singleThreadedWriter` (`ClickHouseBatchWriter`) is initialized to execute synchronously on the CDC thread.
5. **Debezium Engine Instantiation**:
   - `DebeziumEngine.create(Connect.class)` builds the capture engine.
   - Connector class is set (`io.debezium.connector.mysql.MySqlConnector` or `io.debezium.connector.postgresql.PostgresConnector`).
   - Offset backing store is configured (`DebeziumOffsetStorage` or `JdbcOffsetBackingStore` storing offsets in ClickHouse `replica_source_info` with `ORDER BY offset_key`).
   - Engine is dispatched on a dedicated single-threaded executor (`singleThreadDebeziumEventExecutor`).

---

### 3.2 Phase 2: Ingestion Loop & CDC Event Processing

```
Debezium Event Engine
       |
       v (List of ChangeEvent)
handleChangeEventBatch()
       |
       +---> For each record:
       |        |
       |        +---> Extract source timestamp: ClickHouseStruct.getSourceTsFromChangeEvent()
       |        +---> Extract log coordinates: ClickHouseStruct.getSourcePositionFromChangeEvent()
       |        +---> Calculate sequence: nextSequenceNumber(recordTs, position)
       |        |
       |        +---> Check if DDL: isDDLRecord(record)
       |        |        |
       |        |        +--- [YES]:
       |        |        |      1. Flush pending rows: appendToRecords(batch)
       |        |        |      2. drainBeforeDDL() (Queue drain -> Pause -> Await Quiescent)
       |        |        |      3. Parse & Execute DDL on ClickHouse
       |        |        |      4. Invalidate caches: CacheInvalidationManager.invalidateTable()
       |        |        |      5. Acknowledge DDL offset
       |        |        |      6. Resume executor: executor.resume()
       |        |        |
       |        |        +--- [NO]:
       |        |               1. Postgres drift check: checkAndReconcile()
       |        |               2. Parse struct: debeziumRecordParserService.parse()
       |        |               3. If DML: assign sequenceNumber, add to batch
       |        |               4. If Control/Heartbeat (null struct): track as lastControlRecord
       |        |
       +---> Post-loop:
                1. If batch has rows: appendToRecords(batch)
                2. If lastControlRecord != null: commitControlRecordOffset() (if pipeline quiescent)
```

#### Step 2.1: Monotonic Sequence Number & Version Calculation
The connector must assign each change event a strictly monotonic 64-bit `_version` for ClickHouse `ReplacingMergeTree`.
- **Method**: `DebeziumChangeEventCapture.nextSequenceNumber(long recordTs, SourcePosition position)`:
  - `recordTs` is the source commit timestamp (`source.ts_ms`).
  - `SourcePosition` tracks binary log coordinates:
    - MySQL: `(filePrefix, fileSequence, position, row)`.
    - PostgreSQL: `lsn`.
  - **First Delivery vs. Redelivery Logic**:
    - If `position ranks above sequenceHighWaterPosition`, the event is a **first delivery** (new commit).
    - In MySQL, `source.ts_ms` is the statement execution timestamp, not the commit timestamp. A long transaction committing after short transactions carries an older `source.ts_ms`. To maintain commit monotonicity:
      - `effectiveTs` is clamped to the high-water floor: `if (effectiveTs is less than sequenceMaxSourceTs) effectiveTs = sequenceMaxSourceTs`.
      - `sequenceMaxSourceTs` is updated to `effectiveTs`.
    - If `position ranks at or below sequenceHighWaterPosition`, the event is an **offset rewind / redelivery**. The redelivery-stable assignment is preserved without advancing the high-water mark.
  - **Intra-second Sequence Counter**:
    - `diff = (effectiveTs - sequenceAnchorTs) / 1000`.
    - If `diff exceeds 1` (more than 1 second advance): `sequenceAnchorTs = effectiveTs`, `sequenceNumber = SEQUENCE_START (1,000,000,000)`.
    - Else: `sequenceNumber++`.
  - **Resulting Version**: `_version = effectiveTs * 1_000_000L + sequenceNumber`.

#### Step 2.2: DDL Interception & Pre-DDL Inversion Prevention
If an event carries a DDL schema change:
1. **Flush Buffered Rows**:
   - Any DML records accumulated in `batch` prior to the DDL record were captured against the pre-DDL schema.
   - They must be dispatched immediately: `appendToRecords(batch, config)` and `batch.clear()`. This prevents schema inversion where pre-DDL records are executed against a post-DDL table.
2. **Execute DDL Drain Protocol (`drainBeforeDDL()`)**:
   - **Step A (Queue Drain)**: The thread spins while `records != null && !records.isEmpty()`, sleeping 50ms per iteration up to `DDL_DRAIN_TIMEOUT_MS` (60,000ms). Workers continue executing.
   - **Step B (Pause Workers)**: Calls `executor.pause()`. The monitor lock inside `ClickHouseBatchExecutor` sets `isPaused = true`, causing workers arriving at `beforeExecute()` to park.
   - **Step C (Await In-Flight Quiescence)**: Calls `executor.awaitQuiescent(remainingTimeout)`. Spins until `activeBatches.get() == 0`.
3. **Parse and Execute DDL (`performDDLOperation()`)**:
   - DDL is checked against ignore regexes (`checkIfDDLNeedsToBeIgnored`).
   - `DDLParserFactory.getParser()` selects `MySQLDDLParserService` or `PostgreSQLDDLParserService`.
   - The parser generates equivalent ClickHouse SQL (`clickHouseQuery`).
   - Generated columns in MySQL (`GENERATED ALWAYS AS ...`) are mapped to ClickHouse `DEFAULT (expr)`, NOT `MATERIALIZED`, ensuring the column can store source values while deriving default values when omitted.
   - Query is executed via JDBC on `systemDbConnection`.
4. **Cache Invalidation**:
   - Affected table names are resolved from DDL text or event topic.
   - `CacheInvalidationManager.getInstance().invalidateTable(tableKey)` increments the table version counter.
   - If the table name cannot be parsed, `invalidateAll()` increments `globalEpoch` to invalidate all tables.
   - If PostgreSQL, `PostgresSchemaChangeDetector.invalidateCache(tableKey)` clears the schema drift cache.
5. **Resume Pipeline**:
   - Offset is acknowledged via `DebeziumOffsetManagement.acknowledgeRecords()`.
   - `executor.resume()` unlocks the gate and unparks worker threads.

#### Step 2.3: PostgreSQL Schema Drift Detection
For PostgreSQL sources (which do not emit DDL statements over logical replication):
- On every DML record, `pgConfig.getSchemaChangeDetector().checkAndReconcile(sr, dmlTable, dmlDatabase)` is called.
- Compares the `SourceRecord` Kafka Connect `Schema` fields against ClickHouse `system.columns` (cached with TTL).
- If new fields exist in the source record, `PostgresSchemaReconciler.addMissingColumns()` executes `ALTER TABLE db.table ADD COLUMN IF NOT EXISTS col type` on ClickHouse before row insertion proceeds.

#### Step 2.4: Control Record & Heartbeat Offset Commit
- When Debezium completes an initial snapshot, the completion state (`snapshot=false`) is emitted on a subsequent **heartbeat record** (which produces a null `ClickHouseStruct`).
- `commitControlRecordOffset()` checks `isPipelineQuiescent()`:
  - Confirms `records.isEmpty()`, `routedRecords.isEmpty()`, and `DebeziumOffsetManagement.hasUnwrittenBatches() == false`.
  - If quiescent, calls `DebeziumOffsetManagement.acknowledgeRecords()` under `OFFSET_COMMIT_LOCK`. This commits the snapshot offset to ClickHouse `replica_source_info`, preventing snapshot loops on restart.

---

### 3.3 Phase 3: Worker Batch Execution & ClickHouse Insertion

```
ClickHouseBatchRunnable / ClickHouseBatchWriter
       |
       v (poll batch from records queue)
processBatch()
       |
       +---> Register batch: DebeziumOffsetManagement.addToBatchTimestamps(batch)
       |
       +---> Partition records by topic: topicToRecordsMap
       |
       +---> For each topic:
                |
                +---> Resolve Table & Database (apply overrides, prefixes, templates)
                |
                +---> Get or Rebuild DbWriter: getDbWriterForTable()
                |        |
                |        +---> Check CacheInvalidationManager.getVersion()
                |        +---> If stale: re-read schema, rebuild columnNameToDataTypeMap
                |        +---> If table missing & auto.create.tables: autoCreateTable()
                |
                +---> Formulate Queries: GroupInsertQueryWithBatchRecords
                |        |
                |        +---> Staleness check: refreshIfRecordHasUnknownColumn()
                |        |        +---> If unknown col found: re-read system.columns
                |        |        +---> If MATERIALIZED: enforceSourceColumnIsWritable()
                |        |        +---> If unresolvable: markColumnProvenAbsent()
                |        |
                |        +---> Partition by CDC Operation:
                |                 +---> INSERT: After-image fields
                |                 +---> UPDATE: Before-image + After-image fields
                |                 +---> DELETE: Before-image with delete flag
                |                 +---> TRUNCATE: TRUNCATE TABLE
                |
                +---> Execute Batches: flushRecordsToClickHouse()
                |        |
                |        +---> QueryFormatter builds: INSERT INTO table(cols) VALUES (?,?,...)
                |        +---> PreparedStatementExecutor creates JDBC PreparedStatement
                |        +---> PreparedStatementFieldMapper binds types via ClickHouseDataTypeMapper
                |        |        +---> Detect PK relocation: insertTombstonePreparedStatement()
                |        |        +---> Bind _version, _sign, is_deleted
                |        +---> ps.executeBatch()
                |
                +---> Acknowledge Batch: checkIfBatchCanBeCommitted()
                         |
                         +---> Synchronized under OFFSET_COMMIT_LOCK:
                                  recordCommitter.markProcessed()
                                  recordCommitter.markBatchFinished()
```

#### Step 3.1: Batch Dequeue & Tracking
- Worker threads in `ClickHouseBatchRunnable.runLegacyMode()` poll batches from `records`.
- `DebeziumOffsetManagement.addToBatchTimestamps(batch)` calculates the min and max timestamp of the batch and registers it in `inFlightBatches`.

#### Step 3.2: Topic Partitioning & Destination Resolution
- The batch is grouped by Kafka topic: `Map[String, List[ClickHouseStruct]] topicToRecordsMap`.
- Database name is resolved:
  - Source database extracted from record.
  - `clickhouse.database.override.map` applies destination remapping.
  - `clickhouse.common.database.prefix` applies database prefixing.
  - `clickhouse.database.schema.suffix` applies schema template formatting.

#### Step 3.3: DbWriter Lifecycle & Schema Invalidation
- `getDbWriterForTable()` checks the cached `DbWriter` version against `CacheInvalidationManager.getInstance().getVersion(tableKey)`.
- If the table was invalidated by a DDL event:
  - Old `DbWriter` is discarded.
  - Fresh `DbWriter` is instantiated, querying ClickHouse `system.columns` to populate `columnNameToDataTypeMap`.
  - Table engine is detected (`TABLE_ENGINE.REPLACING_MERGE_TREE`, `REPLICATED_REPLACING_MERGE_TREE`, or `COLLAPSING_MERGE_TREE`).
  - Sorting keys are fetched: `DBMetadata.getSortingKeyColumns()`.

#### Step 3.4: Query Template Grouping & Column Enforcement
In `GroupInsertQueryWithBatchRecords.groupQueryWithRecords()`:
1. **Cache Staleness & Unknown Column Probing (`refreshIfRecordHasUnknownColumn`)**:
   - Inspects incoming record schema fields against `cached` ClickHouse columns.
   - If a field is missing from `cached`:
     - Checks `CacheInvalidationManager.isColumnProvenAbsent()`: if proven absent at current version, skips re-reading (prevents metadata query storms).
     - If not proven absent: re-queries `system.columns`.
     - If column appeared: invalidates table version so all workers refresh.
     - If column did NOT appear (e.g. ClickHouse `MATERIALIZED` column):
       - Invokes `enforceSourceColumnIsWritable()`:
         - Inspects `default_kind`. If `ALIAS`, ignored (query-time only).
         - If `MATERIALIZED`, ClickHouse derives it locally and rejects INSERTs. The connector enforces source authority by executing:
           `ALTER TABLE db.table MODIFY COLUMN col (type) DEFAULT (expr)`
         - Re-reads metadata and invalidates cache if successful.
       - If enforcement fails or is skipped, records `markColumnProvenAbsent(tableKey, col)`.
2. **Field Membership Filtering (`QueryFormatter.createColumns`)**:
   - Columns omitted from the record schema (e.g. pre-ALTER records) are excluded from the `INSERT` column list so ClickHouse applies the column `DEFAULT` instead of binding `NULL`.
   - Columns present in the record schema but holding `null` values are **retained** in the `INSERT` list so an explicit `NULL` is bound, preventing ClickHouse from falling back to column defaults.
3. **CDC Operation Handling**:
   - **INSERT (`c`, `r`)**: Dispatches `after` image to `updateQueryToRecordsMap`.
   - **UPDATE (`u`)**:
     - Standard Mode: Generates two records — a before-image record and an after-image record.
     - Replication History Mode: Preserves single after-image record with temporal metadata.
   - **DELETE (`d`)**: Dispatches `before` image with delete flag.
   - **TRUNCATE (`t`)**: Directly generates `TRUNCATE TABLE \`table\``.

#### Step 3.5: Statement Parameter Binding & Execution
In `PreparedStatementExecutor.insertBatch()` and `PreparedStatementFieldMapper.insertPreparedStatement()`:
1. **Primary/Sorting Key Relocation on UPDATE**:
   - If an UPDATE modifies a column that is part of the ClickHouse `ORDER BY` sorting key:
     - `updateRelocatesSortingKey()` detects the mismatch between `beforeStruct` and `afterStruct`.
     - An explicit tombstone row is inserted for the old sorting key:
       `insertTombstonePreparedStatement()` binds the before-image values with `is_deleted = 1` and `_version = record.getVersion() - 1`.
     - The live after-image row is then inserted for the new sorting key with `is_deleted = 0` and `_version = record.getVersion()`.
2. **Type Conversion (`ClickHouseDataTypeMapper.convert`)**:
   - Binds values to `PreparedStatement` parameter positions:
     - Integers: `INT8` (`Byte`), `INT16` (`Short`), `INT32` (`Integer`), `INT64` (`Long`). Uses `Number.intValue()` to prevent class-cast failures on tinyints.
     - Floats/Doubles: `Float32`, `Float64` (converted to `BigDecimal`).
     - Decimals: Scaled `BigDecimal` mapped to `Decimal(p, s)`.
     - Temporals:
       - `MicroTimestamp` / `Timestamp`: Formatted with source/server timezone conversion.
       - `ZonedTimestamp`: Converted using UTC/server zone mapping.
       - `Date` / `Date32`: Converted from epoch days.
       - `MicroTime`: Converted to microseconds since midnight (`Int64`).
     - Strings / JSON: Mapped to ClickHouse `String` or `JSON`.
     - Byte Arrays / Bits: Reverses byte order for Debezium `io.debezium.data.Bits` to match big-endian ClickHouse `Bit` representations.
     - Geometry / Spatial: Uses JTS `WKBReader` to parse WKB polygons and binds to ClickHouse Geo types.
     - Arrays: If ClickHouse target is non-Array, serializes to JSON string; otherwise binds array elements.
3. **System Column Binding**:
   - `_version`: Bound from `record.getVersion()` (`rejectUnderivableVersion` throws if `-1`).
   - `is_deleted`: Bound to `1` for DELETE / tombstones, `0` for live rows.
   - `_sign`: Bound to `-1` for deletes, `1` for inserts (CollapsingMergeTree).
4. **JDBC Execution & Error Classification**:
   - `ps.addBatch()` accumulates records up to `buffer.max.records`.
   - `ps.executeBatch()` flushes to ClickHouse.
   - Errors are passed to `ClickHouseErrorClassifier`:
     - `FATAL` (syntax errors, illegal column types): Batch is discarded and task stops.
     - `UNKNOWN` / `RETRIABLE` (network timeouts, connection loss): Retried on next execution cycle.
     - Special Check: If the exception contains: OffsetStorageWriter is already flushing, the task immediately terminates because Debezium internal offset semaphore has leaked permanently.

#### Step 3.6: Offset Acknowledgement & Commit Serialization
In `DebeziumOffsetManagement.checkIfBatchCanBeCommitted()` and `acknowledgeRecords()`:
1. Verifies no older overlapping batches are still in flight.
2. Acquires `synchronized (OFFSET_COMMIT_LOCK)`.
3. Iterates over records in the batch:
   - Invokes `recordCommitter.markProcessed(record.getSourceRecord())`.
   - On the final record of the batch, invokes `recordCommitter.markBatchFinished()`.
4. Cascades commit checks across `completedBatches` to acknowledge any completed batches that were waiting on this batch.

---

## 4. Method-by-Method Operation Mapping

The following catalog specifies the behavioral contract, synchronization boundaries, and state mutations of the critical operational methods.

#### `public void setup(Properties props, DebeziumRecordParserService debeziumRecordParserService, boolean forceStart)`
- **Purpose**: Initializes and starts the embedded Debezium replication runtime.
- **Inputs**: `props` (configuration properties), `debeziumRecordParserService` (record parser), `forceStart` (flag).
- **Outputs**: None.
- **Preconditions**: ClickHouse database credentials must be accessible; driver must be loadable.
- **Locks**: Thread-confined during startup.
- **Mutations**: Instantiates `records` queue; creates `ClickHouseBatchExecutor`; registers Debezium callbacks; starts background engine thread.
- **Failure Modes**: Throws `ClassNotFoundException` or `IOException`; retries startup up to `MAX_RETRIES` (10) with 10s backoff.

#### `void handleChangeEventBatch(List[ChangeEvent] list, RecordCommitter committer, Properties props, DebeziumRecordParserService parser, ClickHouseSinkConnectorConfig config)`
- **Purpose**: Central CDC dispatch loop invoked by Debezium ChangeConsumer.
- **Inputs**: List of raw change events, batch record committer, configuration.
- **Outputs**: None.
- **Preconditions**: Debezium engine is running and emitting batches.
- **Locks**: Invoked on Debezium event thread. Acquires monitor locks in `nextSequenceNumber` and `drainBeforeDDL`.
- **Mutations**: Flushes pre-DDL rows via `appendToRecords`; triggers `drainBeforeDDL()` on DDL events; tracks `lastControlRecord`; pushes DML batches to `records` queue; invokes `commitControlRecordOffset()`.
- **Failure Modes**: Throws `InterruptedException` if interrupted during queue handoff.

#### `static synchronized long nextSequenceNumber(long recordTs, SourcePosition position)`
- **Purpose**: Generates a strictly monotonic 64-bit version for `ReplacingMergeTree`.
- **Inputs**: `recordTs` (`source.ts_ms`), `position` (`SourcePosition`).
- **Outputs**: `long` version (`effectiveTs * 1_000_000L + sequenceNumber`).
- **Locks**: `synchronized (DebeziumChangeEventCapture.class)`.
- **Mutations**: Updates `sequenceAnchorTs`, `sequenceNumber`, `sequenceHighWaterPosition`, and `sequenceMaxSourceTs`.
- **Failure Modes**: Arithmetic overflow if sequence exceeds 6 digits within 1 second.

#### `private void drainBeforeDDL()`
- **Purpose**: Enforces pipeline quiescence before executing a DDL statement to prevent schema inversion.
- **Inputs**: None.
- **Outputs**: None.
- **Preconditions**: Called on Debezium event thread upon encountering a DDL record.
- **Locks**: Calls `executor.pause()` (acquires `gate`) and `executor.awaitQuiescent()` (acquires `gate`).
- **Mutations**: Transitions executor state to paused; waits for in-flight tasks to terminate.
- **Failure Modes**: Throws `IllegalStateException` if `records` queue does not empty within 60s or executor fails to quiesce.

#### `boolean commitControlRecordOffset(ChangeEvent controlRecord, RecordCommitter committer, boolean handedOffRows)`
- **Purpose**: Commits the offset of non-DML control records (heartbeats) when the pipeline is idle.
- **Inputs**: `controlRecord`, `committer`, `handedOffRows` (flag if current batch produced DML rows).
- **Outputs**: `boolean` (true if committed).
- **Preconditions**: Pipeline must be quiescent (`isPipelineQuiescent() == true`) and `handedOffRows == false`.
- **Locks**: Acquires `OFFSET_COMMIT_LOCK` via `DebeziumOffsetManagement.acknowledgeRecords`.
- **Mutations**: Advances stored offset in `replica_source_info`.
- **Failure Modes**: Returns false without committing if any rows or batches are in flight.

---

#### `public void pause()`
- **Purpose**: Signals executor threads to park before starting new tasks.
- **Inputs**: None.
- **Outputs**: None.
- **Locks**: `synchronized (gate)`.
- **Mutations**: Sets `isPaused = true`; calls `gate.notifyAll()`.

#### `public void beforeExecute(Thread t, Runnable r)`
- **Purpose**: Hook executed by worker threads before running a task body.
- **Inputs**: Worker thread `t`, task `r`.
- **Outputs**: None.
- **Locks**: `synchronized (gate)`.
- **Mutations**: While `isPaused` is true, calls `gate.wait(100)`. Increments `activeBatches`.

#### `public void afterExecute(Runnable r, Throwable t)`
- **Purpose**: Hook executed by worker threads immediately after task completion.
- **Inputs**: Task `r`, exception `t`.
- **Outputs**: None.
- **Locks**: `synchronized (gate)`.
- **Mutations**: Decrements `activeBatches`; calls `gate.notifyAll()`.

#### `public boolean awaitQuiescent(long timeoutMs)`
- **Purpose**: Waits until all active worker tasks have exited their task bodies.
- **Inputs**: `timeoutMs` (maximum wait time).
- **Outputs**: `boolean` (true if `activeBatches == 0`).
- **Locks**: `synchronized (gate)`.
- **Mutations**: Calls `gate.wait()` until `activeBatches.get() == 0` or timeout elapses.

#### `public void resume()`
- **Purpose**: Unparks worker threads and re-enables task execution.
- **Inputs**: None.
- **Outputs**: None.
- **Locks**: `synchronized (gate)`.
- **Mutations**: Sets `isPaused = false`; calls `gate.notifyAll()`.

---

#### `public boolean groupQueryWithRecords(List[ClickHouseStruct] records, Map[QueryTemplate, List[ClickHouseStruct]] queryToRecordsMap, Map[TopicPartition, Long] partitionToOffsetMap, ClickHouseSinkConnectorConfig config, String tableName, String databaseName, Connection connection, Map[String, String] columnNameToDataTypeMap)`
- **Purpose**: Deconstructs record batches into parameterized SQL query templates and associated record buckets.
- **Inputs**: Input records, target query-to-record map, offset tracker, config, table/db names, JDBC connection, cached column types.
- **Outputs**: `boolean` (success status).
- **Locks**: None (thread-confined to worker).
- **Mutations**: Updates `queryToRecordsMap` and `partitionToOffsetMap`; may alter table schema or update cache via `refreshIfRecordHasUnknownColumn`.
- **Failure Modes**: Throws `StaleSchemaCacheException` if record and ClickHouse schema conflict irreconcilably.

#### `private Map[String, String] refreshIfRecordHasUnknownColumn(ClickHouseStruct record, Map[String, String] cached, String tableName, String databaseName, Connection connection, ClickHouseSinkConnectorConfig config)`
- **Purpose**: Detects when incoming records carry columns missing from the local metadata cache; resolves staleness or enforces column writability.
- **Inputs**: Record, cached column map, table/database names, connection, config.
- **Outputs**: Updated column map, or null if cache is valid.
- **Locks**: Reads/writes `CacheInvalidationManager` concurrent maps.
- **Mutations**: Executes `system.columns` queries; executes `ALTER TABLE` if `MATERIALIZED` column needs conversion to `DEFAULT`; marks proven absent columns.
- **Failure Modes**: Returns null on JDBC query failure, falling back to cached map.

#### `private boolean enforceSourceColumnIsWritable(String columnName, String tableName, String databaseName, String fullyQualifiedTableName, Connection connection, ClickHouseSinkConnectorConfig config)`
- **Purpose**: Enforces source-of-truth authority by converting ClickHouse `MATERIALIZED` columns to writable `DEFAULT` columns when the source database supplies values for them.
- **Inputs**: Column name, table name, database name, fully qualified name, connection, config.
- **Outputs**: `boolean` (true if column was successfully converted and is now writable).
- **Locks**: None.
- **Mutations**: Queries `default_kind` and `default_expression` from `system.columns`. Emits `ALTER TABLE db.table MODIFY COLUMN col type DEFAULT (expr)` to ClickHouse.
- **Failure Modes**: Returns false if column is `ALIAS`, expression cannot be read, or DDL execution fails.

---

#### `public void insertPreparedStatement(Map[String, Integer] columnNameToIndexMap, PreparedStatement ps, List[Field] fields, ClickHouseStruct record, Struct struct, boolean beforeSection, ClickHouseSinkConnectorConfig config, Map[String, String] columnNameToDataTypeMap, DBMetadata.TABLE_ENGINE engine, String tableName)`
- **Purpose**: Binds values from a CDC struct to the positional parameters of a JDBC `PreparedStatement`.
- **Inputs**: Column index map, statement `ps`, schema fields, CDC record, data struct, `beforeSection` flag, config, column types, engine type, table name.
- **Outputs**: None.
- **Preconditions**: `ps` must be initialized with placeholders matching `columnNameToIndexMap`.
- **Locks**: None (thread-confined).
- **Mutations**: Sets typed values on `ps` via `setObject()`, `setString()`, `setInt()`, `setNull()`.
- **Failure Modes**: Throws `StaleSchemaCacheException` if a column exists in ClickHouse and the record but has no index placeholder; throws `SQLException` on type binding error.

#### `public void insertTombstonePreparedStatement(Map[String, Integer] columnNameToIndexMap, PreparedStatement ps, List[Field] fields, ClickHouseStruct record, Struct struct, ClickHouseSinkConnectorConfig config, Map[String, String] columnNameToDataTypeMap, DBMetadata.TABLE_ENGINE engine, String tableName)`
- **Purpose**: Binds an explicit `ReplacingMergeTree` delete tombstone row for an UPDATE that changes the row sorting key.
- **Inputs**: Column index map, statement `ps`, before-image fields, record, before struct, config, column types, engine, table name.
- **Outputs**: None.
- **Mutations**: Calls `insertPreparedStatement` with `beforeSection = true`. Overrides delete column to `1` (or `-1`) and overrides version column to `record.getVersion() - 1`.

---

#### `public static synchronized boolean checkIfBatchCanBeCommitted(List[ClickHouseStruct] batch)`
- **Purpose**: Enforces FIFO in-order offset commitment across concurrently executing worker threads.
- **Inputs**: Batch of records that completed JDBC execution.
- **Outputs**: `boolean` (true if batch was committed; false if buffered).
- **Locks**: `synchronized (DebeziumOffsetManagement.class)`.
- **Mutations**: If older overlapping batches exist in `inFlightBatches`, moves batch to `completedBatches`. Otherwise calls `acknowledgeRecords(batch)` and drains qualifying `completedBatches`.

#### `public static void acknowledgeRecord(RecordCommitter committer, SourceRecord record, boolean isLastInBatch)`
- **Purpose**: Thread-safe entry point for committing an individual record offset through Debezium.
- **Inputs**: Debezium `committer`, Kafka `record`, `isLastInBatch` flag.
- **Outputs**: None.
- **Locks**: `synchronized (OFFSET_COMMIT_LOCK)`.
- **Mutations**: Invokes `committer.markProcessed(record)`. If `isLastInBatch` is true, invokes `committer.markBatchFinished()`.
- **Failure Modes**: Throws `RuntimeException` on `InterruptedException`.

---

## 5. Architectural Critique: Incoherencies & Complexity Traps

A rigorous analysis of the 2.11.0 codebase reveals that recurring operational bugs (stuck offsets, silent data divergence, snapshot loops, metadata query storms, and DDL drain deadlocks) are not isolated defects, but direct consequences of fundamental architectural mismatches.

### 5.1 Impedance Mismatch: Relational In-Place Mutation vs. Columnar Append-Only Storage
- **The Core Conflict**: Relational CDC engines assume mutable, in-place storage with transactional read isolation. ClickHouse is an append-only columnar database where row updates and deletions do not exist at the storage layer; instead, mutations are simulated via merge-tree engines (`ReplacingMergeTree`, `CollapsingMergeTree`).
- **The Resulting Complexity**:
  - The connector must synthesize artificial columns (`_version`, `_sign`, `is_deleted`) and insert tombstone rows to represent deletes.
  - When an UPDATE modifies any column belonging to the ClickHouse `ORDER BY` sorting key, ClickHouse `ReplacingMergeTree` cannot replace the old row (since the sorting key differs). The connector is forced to synthesize an artificial tombstone for the old key and an insert for the new key.
  - Queries against the replica observe duplicate and un-deleted rows unless `FINAL` or `argMax` is explicitly queried, shifting the burden of consistency onto the query layer.

### 5.2 The Arithmetic Version Bleed & Restart Inversion Defect
- **The Formula**: `_version = ts_ms * 1_000_000L + sequenceNumber`.
- **The Structural Flaw**:
  - `ts_ms` is multiplied by 10^6, allocating 6 decimal digits for `sequenceNumber`.
  - However, `SEQUENCE_START` is initialized to `1_000_000_000` (10 decimal digits), and on restart `SEQUENCE_START_INITIAL` is set to `500_000_000` (9 decimal digits).
  - The addition of a 9- or 10-digit sequence number **bleeds directly into the timestamp digits**, shifting the effective timestamp by approximately 500 to 1000 seconds.
- **The Restart Inversion Failure Mode**:
  - When the connector restarts, `SEQUENCE_START_INITIAL` (500m) is chosen so that replayed records receive lower versions than their pre-restart writes (seeded at 1000m).
  - However, if a *genuinely newer* transaction commits at T + 1 ms immediately after restart, its version calculation produces:
    (T + 1) * 1e6 + 500,000,000 is strictly less than T * 1e6 + 1,000,000,000
  - The newer event receives a **strictly lower version** than the pre-restart event for the same key. ClickHouse `ReplacingMergeTree` permanently discards the newer write, causing silent data loss.

### 5.3 Global Static State Anti-Pattern & Multi-Tenancy Hazards
- Critical coordination state across the codebase is held in static variables:
  - `DebeziumChangeEventCapture.sequenceNumber`, `sequenceAnchorTs`, `sequenceHighWaterPosition`, `sequenceMaxSourceTs`.
  - `DebeziumOffsetManagement.inFlightBatches`, `completedBatches`, `outstandingBatches`, `OFFSET_COMMIT_LOCK`.
  - `CacheInvalidationManager.INSTANCE`.
- **Consequences**:
  - Multiple connectors or tasks running within the same JVM cross-contaminate state.
  - Thread pools share singletons across independent pipelines, leading to lock contention and corrupted high-water marks.

### 5.4 DDL Synchronization Fragility & Thread Pool Deadlocks
- **The Issue**: Synchronizing synchronous DDL events across asynchronous worker threads via `drainBeforeDDL()` is inherently fragile.
- **Historical Deadlock**: Pausing the executor before draining the queue caused worker threads to park in `beforeExecute()`, preventing any dequeuing and guaranteeing a 60s timeout abort that discarded in-flight batches.
- **Current Vulnerability**: The drain loop relies on a fixed 60,000ms deadline. Under heavy ClickHouse write pressure, batch inserts take longer to flush, causing `drainBeforeDDL()` to time out and abort the DDL, triggering continuous restart loops.

### 5.5 Metadata Cache Storms & Tri-Partite Schema Engines
- The repository maintains three separate, disconnected schema evolution engines:
  1. MySQL ANTLR DDL parser (`MySqlDDLParserService`) in lightweight mode.
  2. PostgreSQL schema drift detector (`PostgresSchemaChangeDetector`) in lightweight mode.
  3. Kafka Connect `ClickHouseAlterTable` in Kafka mode.
- In lightweight mode, `refreshIfRecordHasUnknownColumn` was previously unable to distinguish between a stale schema cache and ClickHouse-computed columns (`MATERIALIZED`/`ALIAS`). This drove an unbounded metadata loop (observed at 9.66 million `system.columns` queries/hour) that exhausted ClickHouse connection pools and stalled replication.

### 5.6 Offset Storage Semaphore Leaks & Masked Divergence
- In Debezium `EmbeddedEngine`, `commitOffsets()` leaks the `flushInProgress` semaphore if multiple threads call `markProcessed()` and `beginFlush()` concurrently without global mutual exclusion.
- When this occurred, `ClickHouseErrorClassifier` categorized the resulting `ConnectException: OffsetStorageWriter is already flushing` as an `UNKNOWN` retriable error. With infinite retries, the connector continued inserting data into ClickHouse while offsets remained frozen, allowing replication to drift hours behind without failing loudly.

### 5.7 Code Duplication & Divergence
- `ClickHouseBatchWriter` (single-threaded mode) and `ClickHouseBatchRunnable` (multi-threaded mode) duplicate approximately 1,000 lines of nearly identical pipeline logic (`getDbWriterForTable`, `processRecordsByTopic`, `flushRecordsToClickHouse`).
- Bug fixes applied to one path (such as lock serialization or timezone handling) have repeatedly failed to be mirrored in the other, creating subtle behavioural divergences.
- Hash-based routing (`appendToRecordsWithHashRouting`) remains partially implemented but hardcoded to `useHashRouting = false` in `ClickHouseBatchRunnable`, leaving dead code with busy-spin loops (`Thread.sleep(10)`).

---

## 6. Blueprint for Architectural Simplification

To transition the sink connector from a reactive, bug-patching posture into a robust, high-performance replication engine, the following architectural redesign is recommended at the specification level:

```
+----------------------------------------------------------------------------------------------------+
|                                    PROPOSED TARGET ARCHITECTURE                                    |
+----------------------------------------------------------------------------------------------------+
|                                                                                                    |
|   +--------------------------------------------------------------------------------------------+   |
|   | 1. DETERMINISTIC LOG-BASED VERSIONING                                                      |   |
|   |    - Replace synthetic timestamp arithmetic with 64-bit monotonic log position:            |   |
|   |      UInt64 = bitShiftLeft(BinlogFileSeq, 32) | BinlogOffset  (MySQL)                      |   |
|   |      UInt64 = LSN (PostgreSQL)                                                             |   |
|   |    - Zero arithmetic carry; zero timestamp drift; 100% deterministic across restarts.       |   |
|   +--------------------------------------------------------------------------------------------+   |
|                                                                                                    |
|   +--------------------------------------------------------------------------------------------+   |
|   | 2. BARRIER-BASED DDL SYNCHRONIZATION                                                       |   |
|   |    - Eliminate thread pausing and timeout-based drain loops.                               |   |
|   |    - Insert a special DdlBarrierRecord into the internal worker queue.                     |   |
|   |    - Workers process all preceding batches; the barrier executes DDL exclusively once     |   |
|   |      preceding batches complete, then resumes normal flow. Deadlock-free by construction.  |   |
|   +--------------------------------------------------------------------------------------------+   |
|                                                                                                    |
|   +--------------------------------------------------------------------------------------------+   |
|   | 3. UNIFIED INGESTION ENGINE & REMOVAL OF DUPLICATION                                       |   |
|   |    - Eliminate ClickHouseBatchWriter; converge all execution onto ClickHouseBatchRunnable.  |   |
|   |    - Unify Kafka Connect and Lightweight modes to share an identical batching core.        |   |
|   |    - Remove all static mutable singletons; encapsulate state inside task-scoped contexts.  |   |
|   +--------------------------------------------------------------------------------------------+   |
|                                                                                                    |
|   +--------------------------------------------------------------------------------------------+   |
|   | 4. STRICT SOURCE-OF-TRUTH SCHEMA SYNCHRONIZATION                                           |   |
|   |    - Consolidate MySQL and PostgreSQL schema evolution into a single Catalog Manager.       |   |
|   |    - Maintain an explicit column contract: ordinary columns mirror source; ALIAS is        |   |
|   |      ignored; MATERIALIZED is strictly prohibited or converted to DEFAULT on startup.       |   |
|   |    - Hard-fail startup if a replicated table lacks a primary/unique key (GIPK mandatory).  |   |
|   +--------------------------------------------------------------------------------------------+   |
|                                                                                                    |
+----------------------------------------------------------------------------------------------------+
```

### Conclusion
By mapping the system to the tee, this specification clarifies every critical path in the 2.11.0 release. The systemic bugs observed in production—offset stalls, version inversions, DDL deadlocks, and metadata storms—stem from shared mutable state, arithmetic version encoding, and decoupled asynchronous DDL execution. Addressing these issues via deterministic log-based versioning, barrier-based DDL synchronization, and pipeline unification will provide a stable, coherent foundation for all future development.


minguyen9988:
CI watch: testflows-lightweight-x86 / testflows-lightweight and testflows-lightweight-x86-hikari-pool / testflows-lightweight failed on 3c658181 (run 35855141814). Triage from the completed sibling run 35835468969 (0ac8224e, same tree minus #1484): the scenarios alter/modify column not null and alter/change column name to new name not null fail on every table variant after the 100 s DESC retry. Root cause: #1479 changed Spec 06.05 3.2 rule 2 so a MODIFY or CHANGE COLUMN ... NOT NULL on an existing non-Nullable ClickHouse column keeps it non-Nullable; the suite tables declare x INT NOT NULL, so the replica column is Int32 and becomes String, but the two scenarios still asserted Nullable(String) and their docstrings described the old always-Nullable rule. Fix commit 5207850a updates the two assertions and docstrings (the plain modify column scenario already asserted String and passed; explicit NULL scenarios still assert Nullable(String)). Verified: python3 -m py_compile on alter.py, scripts/validate_specs.py --changed-base origin/develop PASSED (60 specs, 576 citations), python3 -m unittest discover -s scripts/tests 39 OK. Still open and tracked separately: the _no_primary_key and _no_engine_no_primary_key variants of the alter and compound-alter scenarios fail because the altered column x is part of the keyless all-columns sorting key (Spec 06.05 3.6, Code 524 on MODIFY or RENAME of a key column); their rename, re-type and FIRST/AFTER reposition must go through the Spec 06.09 rebuild path, which does not yet apply column position. That is a product change and will follow spec-first.

sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/api/DebeziumEmbeddedRestApi.java
```diff
@@ -331,6 +381,11 @@ public static void startRestApi(Properties props, Injector injector,
     /**
      * Stops the Javalin REST API server.
      */
+    /** A JSON field as a string: numbers are accepted for positions, null stays null. */
```
subkanthi: maybe this can be moved to utils so its universally accessible

sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DdlCaptureFilter.java
subkanthi: This class has to be in the ddl package

subkanthi (changes requested):

sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DdlCaptureFilter.java
```diff
@@ -0,0 +1,104 @@
...
+     *                 statement (e.g. {@code CREATE DATABASE}).
+     * @param props    the connector properties.
+     * @return true when the DDL belongs to a captured table (or, for a
+     *         table-less DDL, a captured database).
+     */
+    static boolean isCaptured(String database, String table, Properties props) {
+        if (database == null || database.isEmpty()) {
+            return true;
+        }
+        if (!passes(database, props.getProperty("database.include.list"),
```
subkanthi: please dont use hardcoded debezium variables, when debezium changes it becomes hard here, `database.exclude.list`, should be accessible from Debezium library.

sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DdlCaptureFilter.java
```diff
@@ -0,0 +1,104 @@
...
+                props.getProperty("database.exclude.list"))) {
```
subkanthi: please dont use hardcoded debezium variables, when debezium changes it becomes hard here, `database.exclude.list`, should be accessible from Debezium library.

sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/cdc/DdlCaptureFilter.java
```diff
@@ -0,0 +1,104 @@
...
+            return false;
+        }
+        if (table == null || table.isEmpty()) {
+            return true;
+        }
+        return passes(database + "." + table, props.getProperty("table.include.list"),
```
subkanthi: please dont use hardcoded debezium variables, when debezium changes it becomes hard here, `table.exclude.list`, should be accessible from Debezium library.

minguyen9988:
CI watcher: testflows-lightweight-x86 and testflows-lightweight-x86-hikari-pool (non-replicated) fail on the keyless-table variants of the alter suite (modify_column_null, modify_column_first_after, change_column_name_to_new_name_null, change_column_first_after and the compound add/change and add/modify alters). Root cause, three defects in the sorting-key policy for tables keyed on every stored column (Spec 06.05 3.6, so x is a key column): (1) MODIFY x INT FIRST was suppressed as a same-type restatement, dropping the reorder, although ClickHouse accepts a MODIFY that restates the identical type with FIRST/AFTER on a key column (metadata-only, measured on 24.8.14); (2) CHANGE x x2 INT FIRST deferred the rename to the rebuilt table but lost the position; (3) MODIFY x VARCHAR(100) NULL deferred Nullable(String) but the same-identity rebuild stripped the Nullable from re-typed key columns unless the plan was the keyless fallback, so the rebuilt column could not hold the NULLs the source now sends. Fix commit ea4755d5: the translator restates the existing type with the position (Spec 06.05 3.4 rule 2), the rebuild plan carries positionedColumns and step 3b restates the column with its rebuilt-table type and position on the empty rebuilt table, and a Nullable re-type is kept with allow_nullable_key = 1 (Spec 06.09 3.1.1, 3.3.1). Verified on JDK 17: MySqlDDLParserListenerImplTest 156 tests and PrimaryKeyRebuildTest 24 tests green with three new tests, the whole ddl unit package 261 tests green, validate_specs.py --changed-base origin/develop PASSED (60 specs, 578 citations), scripts/tests 39 OK.

minguyen9988:
CI watch: java-tests-lightweight (run 35855141814, job 107163673741, head 3c658181) failed 10 of 14 PrimaryKeyChangeIT scenarios with Code 81 Database employees does not exist on every connector statement after the snapshot DDL DROP DATABASE IF EXISTS employees, including the restarted engine's first query. Root cause: ITCommon.getDBWriter built a jdbc URL ending in /employees but registered the pool as system. Pools are process-wide and keyed by host:port and database (Spec 03.05 section 3.2), so when the test helper ran before the engine's setSystemDbConnection (PrimaryKeyChangeIT.Scenario obtains the writer right after launching the engine thread) the engine inherited a system pool whose connections defaulted to employees, and the replicated DROP DATABASE removed the connection's own default database. The ReplicatedRMT suites ran the same DROP and CREATE DATABASE pair on a real system connection in the same job and passed. Fix a25895e9: the helper now pairs the system pool name with a system URL, as ClickHouseBatchRunnable and ClickHouseBatchWriter already do. Test-only change, spec-exempt. Verified on JDK 17: sink-connector-lightweight test-compile, validate_specs.py --changed-base origin/develop PASSED (60 specs, 578 citations), scripts/tests 39 OK. The remaining 3c658181 failures (AlterTableModifyColumnIT class_name Nullable stripped on the keyless rebuild, already addressed by ea4755d5; TableOperationsIT, IsDeletedColumnsIT, MySQLDemoIT and the DateTime suites reporting missing tables or data) are being classified from the same log next.

minguyen9988:
CI watch, second fix for the same java-tests-lightweight job (run 35855141814, head 3c658181): at 13:53Z the first out-of-range row of temporal_types_DATETIME2 (1000-01-01, from the datetime.sql and data_types.sql fixtures) threw DebeziumConverter ValueOutOfRangeException in the writer, the worker stopped with FATAL ClickHouse error (Code -1) and the engine stopped on the next batch, so DateTimeWithUserProvidedDifferentTimeZoneIT, CreateTableDataTypesTimeZoneIT, DateTimeWithTimeZoneColumnSchemaOnlyIT, DateTimeWithUserProvidedTimeZoneSchemaOnlyIT, TableOperationsIT and IsDeletedColumnsIT (all built on those fixtures and asserting the bounded 1900-01-01 and 2299-12-31 values) failed on missing data. The commit that made an unrepresentable value terminal opted the testflows suites in through tests/integration/helpers/default_config.py and the Postgres ITs through PostgresProperties, but the MySQL ITs load src/test/resources/config.yml, which stayed on the loud default. Fix f099689b adds clamp.out.of.range=true to that config.yml with the same rationale; the product default is unchanged and stays pinned by DebeziumConverterRangePolicyTest and PreparedStatementFieldMapperOutOfRangeTest. Test resource only, spec-exempt. Verified: YAML parses (34 keys), validate_specs.py --changed-base origin/develop PASSED (60 specs, 578 citations), scripts/tests 39 OK.

minguyen9988:
CI watch, fix 8c36472b (pushed to 2.11.0): PrimaryKeyBackfill completeness check failed forever after a key-column re-type. In the java-tests-lightweight log of the 3c658181 run (and in every alter-suite scenario that re-types a keyless table's column), Spec 06.09 step 3 joined the retired table's key column against the rebuilt table's re-typed column with no conversion, so ClickHouse rejected the probe with Code 386 NO_COMMON_TYPE (Left key class_name type Nullable(String), Right key class_name type Int32) on every attempt; the backfill retried with backoff indefinitely and the retired copy was never released. Tests still passed because step 2 (the copy) had already inserted the rows. Fix: the retired side of the probe now reads a re-typed key column through the conversion the copy applied, CAST(r.old, type on T), or ifNull(CAST(r.old, Nullable(type on T)), defaultValueOfTypeName(type on T)) when only R has it Nullable (the value insert_null_as_default stored). Measured on ClickHouse 24.8.14: the old probe raises 386, the new one returns 0 after the copy for Nullable and non-Nullable targets, NULL included. Spec 06.09 3.3.2 step 3 amended; PrimaryKeyRebuildTest.completenessCheckCastsRetypedKeyColumns pins both statement shapes and the conversion's edge cases. Verified: PrimaryKeyRebuildTest 25 run 0 failures on JDK 17, validate_specs.py PASSED (60 specs, 579 citations), scripts/tests 39 OK. Also from that log: the two remaining java failures (AlterTableChangeColumnIT.testChangeColumn and AlterTableModifyColumnIT.testAlterAddPrimaryKeyAndModifyNotNull) were the Nullable-stripping defect fixed in ea4755d5 and its cascade (the first test leaks its engine when an assertion fails, so the second test's engine waited 60 s on the metrics MBean and its 50 s poll expired). On the ea4755d5 run, testflows-lightweight-x86 non-replicated passed (1h55m) before that run was cancelled by the next push.

minguyen9988:
CI watch, java-tests-lightweight (runs 35899616412 on a25895e9 and 35900258649 on f099689b, 17 and 6 failures; the head run on 8c36472b is still in progress). Two fixes pushed to 2.11.0. (1) 345f88f7 fix(pk-rebuild): AlterTableModifyColumnIT.testModifyColumn (ship_class.class_name expected Nullable(Int32) but was Int32), AlterTableChangeColumnIT.testChangeColumn (class_name_new likewise) and the cascading testAlterAddPrimaryKeyAndModifyNotNull. Root cause: ship_class is keyless, so it is keyed on every stored column, all Nullable under allow_nullable_key = 1 (Spec 06.05 3.6); the same-identity rebuild planned for a deferred MODIFY/CHANGE of a key column was flagged as a declared key, and rewriteCreateStatement re-declares every non-retyped key column of a declared key non-Nullable, so the SECOND key-column ALTER on the table (tonange) stripped the Nullable the first rebuild had kept on class_name. With rows, the copy would store type defaults for NULLs in the untouched key columns. Fix: the same-identity plan is flagged keyless when the sorting key is the all-columns identity (Spec 06.09 3.1.2), so every key column keeps its nullability; declared keys unchanged. Verified: MySqlDDLParserListenerImplTest 157 run 0 fail (new testSameIdentityRebuildOfKeylessTableIsFlaggedKeyless), PrimaryKeyRebuildTest 26 run 0 fail (new keylessSameIdentityRebuildKeepsOtherKeyColumnsNullable), validate_specs.py PASSED (60 specs, 581 citations), scripts/tests 39 OK. (2) 95fd86f4 test(it): TableOperationsIT.testTableOperations[1]/[2] compared SHOW CREATE TABLE employees.members against the pre-2.11.0 rendering ORDER BY tuple(); members and rcx are keyless, so the translator now emits ORDER BY (all columns) SETTINGS allow_nullable_key=1 (the shape MySqlDDLParserListenerImplTest already pins for these two statements). Expectations updated to the rendering produced by clickhouse-local 24.8.14 from the translator's exact CREATE statements, and the asserts report the observed statement. Still under investigation from the same log: PrimaryKeyChangeIT.truncateDuringBackfillLeavesTableEmpty (300000 rows remain after the TRUNCATE issued while the backfill was pending); the job log is not downloadable until the run completes.

minguyen9988:
CI triage: java-tests-lightweight (run 35900258649 on f099689b) failed PrimaryKeyChangeIT.truncateDuringBackfillLeavesTableEmpty with 'expected 0 live rows in employees.t_big, found 300000'. Root cause from the job log: the backfill's copy INSERT ... SELECT was issued at 18:42:05.319, the replicated TRUNCATE reached the connector at 05.824, the cancel marked the task and returned at once, the TRUNCATE ran on ClickHouse at 05.832 and returned in 2 ms, and the copy statement only returned before 06.414 - so every copied row landed after the truncate. Spec 06.09 step 7 assumed ClickHouse would order the two behind the table's exclusive lock; it does not: for MergeTree tables TRUNCATE TABLE takes no exclusive table lock (InterpreterDropQuery, Truncate branch, 24.8: 'We don't need any lock for ReplicatedMergeTree and for simple MergeTree'). Fix 4cc10e90: PrimaryKeyBackfill.cancel() now waits for the running attempt to stop at its next statement boundary (the in-flight statement has returned) before dropping the retired copy and returning to the DDL path, so the superseding TRUNCATE/DROP always runs after the last copy statement (uncapped wait, WARN every 30 s, loud on interrupt, no wait when the cancel runs on the attempt thread itself). Spec 06.09 3.3.2 step 7 and section 5 updated in the same commit. Verified: new PrimaryKeyRebuildTest.cancelWaitsForInFlightCopyStatement (fails without the wait at 'the cancel must not return while the copy statement is in flight'), PrimaryKeyRebuildTest 27/0, MySqlDDLParserListenerImplTest 157/0 on JDK 17, validate_specs.py --changed-base origin/develop PASSED (60 specs, 581 citations), scripts/tests 39 OK.

minguyen9988:
Review comments from subkanthi addressed in b8def93e (pushed to 2.11.0), no behaviour change. (1) DdlCaptureFilter now lives in com.altinity.clickhouse.debezium.embedded.ddl (public, since DebeziumChangeEventCapture calls it across packages); DdlCaptureFilterTest moved with it and the spec 06.08 codebase mapping cites the new path. (2) The Debezium list property names are no longer hardcoded: DdlCaptureFilter exposes DATABASE_INCLUDE_LIST, DATABASE_EXCLUDE_LIST, TABLE_INCLUDE_LIST and TABLE_EXCLUDE_LIST read from io.debezium.relational.RelationalDatabaseConnectorConfig.X.name(), and the three other readers this PR had introduced with the same literals (KeylessTablePreflight.withoutExcluded and databaseFilter, VersionHighWaterMark.targetDatabases, the backfill resumption scan in DebeziumChangeEventCapture) go through those constants, so a rename on the Debezium side is caught at the dependency bump instead of silently turning the filter into a no-op. New test DdlCaptureFilterTest.propertyNamesComeFromDebezium pins the link and checks the filter honours a list set under Debezium's own field name; spec 06.08 3.3 rule 3 states the rule. (3) stringOrNull moved from DebeziumEmbeddedRestApi to the shared com.altinity.clickhouse.sink.connector.common.Utils so both modules can use it (the javadoc it had displaced from stop() is back in place), with UtilsTest.testStringOrNull covering null, numeric and string inputs. Verified on JDK 17: both modules compile offline; UtilsTest 5/0; DdlCaptureFilterTest 9/0, DdlIgnoreRulesTest 7/0, KeylessTablePreflightTest 30/0, VersionHighWaterMarkTest 9/0, DebeziumEmbeddedRestApiDoubleStartTest 3/0; scripts/validate_specs.py --changed-base origin/develop PASSED (60 specs, 581 citations); scripts/tests 39 OK.

minguyen9988:
CI: testflows-lightweight-replicated (both the plain and the hikari-pool variant) failed on every run of this branch from the scenario alter / change column first after / no primary key onwards and ran into the 6h job timeout, while the non-replicated variants and the Java ITs passed. Root cause: on a ReplicatedReplacingMergeTree target the primary-key rebuild (Spec 06.09) derived the scratch CREATE from SHOW CREATE TABLE and issued it as a plain query. A table created ON CLUSTER with empty engine arguments renders the default path /clickhouse/tables/{uuid}/{shard}, and ClickHouse accepts the {uuid} macro only in an ON CLUSTER query (registerStorageMergeTree: allow_uuid_macro only on cluster, Replicated database or ATTACH; Code 36 Macro uuid and empty arguments of ReplicatedMergeTree are supported only for ON CLUSTER queries with Atomic database engine). Step 3 of the swap therefore failed on every keyless or deferred-clause key-column ALTER, the DDL barrier threw and nothing was replicated afterwards, which is why every later scenario failed too. Fix 22eae7a6 (spec first, 06.09 3.2 item 3, 3.3.1 Replicated mode, 3.3.2 steps 4 and 7): when auto.create.tables.replicated=true -- the same switch on which the translator appends ON CLUSTER to its own DDL -- every DDL the swap and the backfill issue on the mirrored table and its copies (leftover DROP, CREATE TABLE S with the engine arguments kept verbatim, deferred ALTER TABLE S, the MODIFY COMMENT marker, EXCHANGE TABLES / RENAME TABLE, and the DROP or MODIFY COMMENT of the retired copy on completion and on a superseding TRUNCATE/DROP) carries ON CLUSTER {cluster}; the copy and the completeness check stay on the copying host and the local key map never carries it. Without the switch nothing changes. Verified: new unit tests PrimaryKeyRebuildTest.replicatedSwapRunsOnCluster and replicatedBackfillDropsRetiredOnCluster (both red when either onCluster() overload returns an empty string), PrimaryKeyRebuildTest 29/0 and MySqlDDLParserListenerImplTest 157/0 on JDK 17, validate_specs.py --changed-base origin/develop and scripts/tests pass. The replicated testflows suites cannot run outside CI, so this run is the confirmation.

minguyen9988:
@subkanthi status of each of your review comments, checked against the current head f4cc17af (b8def93e is an ancestor of it). All three points are changed the way you asked; nothing is pending on our side.

1. DebeziumEmbeddedRestApi.java line 384, "maybe this can be moved to utils so its universally accessible" (https://github.com/Altinity/clickhouse-sink-connector/pull/1437#discussion_r4083922321) -- FIXED in b8def93e. stringOrNull now lives in the shared module as com.altinity.clickhouse.sink.connector.common.Utils.stringOrNull (Utils.java line 441) and DebeziumEmbeddedRestApi calls Utils.stringOrNull at lines 209-211 and 305; the private copy is gone and the stop() javadoc it had displaced is back in place. Covered by UtilsTest.testStringOrNull (null, numeric and string inputs).

2. DdlCaptureFilter.java, "This class has to be in the ddl package" (https://github.com/Altinity/clickhouse-sink-connector/pull/1437#discussion_r4083948880) -- FIXED in b8def93e. The class is now sink-connector-lightweight/src/main/java/com/altinity/clickhouse/debezium/embedded/ddl/DdlCaptureFilter.java (public, since DebeziumChangeEventCapture calls it from the cdc package); DdlCaptureFilterTest moved with it and spec 06.08 cites the new path. Nothing is left under cdc/.

3. DdlCaptureFilter.java lines 55, 56 and 62, "please dont use hardcoded debezium variables ... should be accessible from Debezium library" (https://github.com/Altinity/clickhouse-sink-connector/pull/1437#discussion_r4083967612, https://github.com/Altinity/clickhouse-sink-connector/pull/1437#discussion_r4083968612, https://github.com/Altinity/clickhouse-sink-connector/pull/1437#discussion_r4083971179) -- FIXED in b8def93e. The four keys are read from io.debezium.relational.RelationalDatabaseConnectorConfig.DATABASE_INCLUDE_LIST / DATABASE_EXCLUDE_LIST / TABLE_INCLUDE_LIST / TABLE_EXCLUDE_LIST .name() (DdlCaptureFilter lines 47-57) and exposed as constants; isCaptured uses those constants, and the three other readers this branch had introduced with the same literals now go through them as well (KeylessTablePreflight lines 491, 493, 558; VersionHighWaterMark line 312; DebeziumChangeEventCapture line 3462). A grep of both main source trees on the head for the quoted literals "database.include.list", "database.exclude.list", "table.include.list", "table.exclude.list" returns nothing. Pinned by DdlCaptureFilterTest.propertyNamesComeFromDebezium, which asserts the constants equal the Debezium field names and that the filter honours a list set under the Debezium name.

CI on f4cc17af: build, publish, java-tests-lightweight, java-tests-kafka, checksum, Lean proofs and spec validator, and testflows-lightweight (non-replicated) are green; the testflows-lightweight-replicated jobs are still running. When you are happy with the above, please resolve the five threads and re-review so the CHANGES_REQUESTED flag clears.

PR #1441, opened 2026-09-09
fix(ddl): translate a MySQL generated column to DEFAULT, not MATERIALIZED

minguyen9988:


minguyen9988:
Superseded by #1442 — closing in favour of that one.

This PR was opened against `develop`; the base was then retargeted to `2.11.0`, which left it `CONFLICTING`. I cannot rebase the head branch in place (no force-push available to me), so #1442 is the identical three-file change applied cleanly on top of `2.11.0` (`b05b6aa`).

Also worth noting for anyone reading the red check here: the `java-lightweight` failure on this PR was `wget` exiting 4 while downloading Maven 3.8.8 from `archive.apache.org` in the **Install maven 3.8.8** step — a transient network failure on the runner, not a build or test error. No source was compiled and no test ran in that job.

PR #1448, opened 2026-09-13
fix(ddl): the MATERIALIZED-to-writable ALTER is a no-op, and would lose the derivation if it were not

minguyen9988:
## The problem

`DBMetadata#makeColumnWritable()` exists to stop a ClickHouse MATERIALIZED column shadowing the value the source supplies. It emits:

```sql
ALTER TABLE `db`.`t` MODIFY COLUMN `c` <type>
```

on the belief that restating the column without a default clause drops the MATERIALIZED expression. **It does not.** ClickHouse reads an omitted default clause as "leave the existing default alone": the statement is accepted, returns no error, and `default_kind` stays `MATERIALIZED`.

The source value goes on being discarded. The post-ALTER verification added alongside it is the only reason this surfaces at all - it logs `Enforcement did not take effect` once per DDL generation, forever, and the divergence is never corrected.

Verified in an isolated `clickhouse-local` sandbox, reproduced identically on 24.8.14, 25.8.12 and 26.1.6:

```
CREATE TABLE t (id Int64, base Nullable(Int64),
                c Nullable(Int64) MATERIALIZED ifNull(base,0)*2)

ALTER TABLE t MODIFY COLUMN c Nullable(Int64)
    -> default_kind: MATERIALIZED        (unchanged, no error)

ALTER TABLE t MODIFY COLUMN c Nullable(Int64) DEFAULT ifNull(base,0)*2
    -> default_kind: DEFAULT
```

## Why the obvious repair is the wrong one

The tempting fix is `MODIFY COLUMN c REMOVE MATERIALIZED`, which does work. It is worse than leaving the bug in place.

A MATERIALIZED column is **derived**: the replica computes it whenever the connector omits it from the INSERT. A bare column derives nothing. Every row written without that column would store a type zero - and these are precisely the tables whose source declares the column generated, so the connector omits it routinely. That trades a value divergence for a data-loss path.

## The fix

Convert to `DEFAULT` over the same expression. DEFAULT is the only kind that satisfies both halves: it accepts an explicit value, so the binlog value lands as sent and the source becomes authoritative; and it re-derives the value when the column is absent, exactly as MATERIALIZED did. Same sandbox, same builds:

```
INSERT INTO t (id, base)    VALUES (1, 5)        -> c = 10   (derived)
INSERT INTO t (id, base, c) VALUES (2, 5, 999)   -> c = 999  (source wins)
```

This is the same conclusion #1442 reached for the CREATE TABLE path, where a MySQL generated column is now translated to DEFAULT rather than MATERIALIZED. This PR brings the ALTER path into agreement with it. Until now the two disagreed, so a table created before #1442 could never be corrected into the shape #1442 creates.

### Changes

- **`makeColumnWritable`** now emits `MODIFY COLUMN <type> DEFAULT <expr>`. The expression is read back from `system.columns` and restated verbatim rather than reconstructed, so the replica keeps deriving precisely what it derived before. CODEC, COMMENT and column TTL are carried across by ClickHouse itself - verified on the same builds by comparing `create_table_query` either side of the conversion.
- **`getColumnDefaultExpression`** - new read-only introspection, the counterpart of the existing `getColumnType` / `getColumnDefaultKind` pair.
- **Unreadable expression aborts the conversion**, issuing no DDL. Emitting the bare form there is the no-op above, and any statement that did take effect would remove the derivation.
- **The post-ALTER verification asserts the column reads back as `DEFAULT`**, rather than asserting it is no longer MATERIALIZED. The negative form also accepts a column left with no default at all - the outcome this change exists to avoid.
- Caller log messages and javadoc updated: they described *dropping* the expression, which is neither what happens nor what should.

ALIAS behaviour is unchanged (still skipped - it stores nothing, so there is nothing to diverge). Ordinary columns are unaffected. `makeColumnWritable` has exactly one production caller; its signature and return contract are unchanged.

## Testing

Built with JDK 17, maven 3.9.11, on `2.11.0` @ `50cd189a`.

`UnwritableColumnReportingTest`: **16 run, 0 failures.**

Mutation-checked - each mutation is caught by a different test:

| mutation | result |
|---|---|
| emitter reverted to the bare `MODIFY COLUMN` | 3 failures |
| verification reverted to "not MATERIALIZED" | 1 failure |
| unreadable-expression guard disabled | 1 failure |

Full module suite, same tree:

```
unmodified 2.11.0:  Tests run: 372, Failures: 0, Errors: 20
with this change:   Tests run: 376, Failures: 0, Errors: 20
```

The 20 errors are identical in both runs - `Previous attempts to find a Docker environment failed`, from the container-backed tests. Pre-existing and unrelated; no Docker on the build host.


minguyen9988:
Adversarially reviewed by a 3-model cross-provider panel (Gemini 3.1 Pro / Grok 4.2 Reasoning / GPT-5.5); unanimous approval on the second round at head b3769d1f.

The second commit is that review output. Two findings were real and were introduced by the first commit, both in the new lookup: the identifiers were interpolated into SQL literals (a replicated name containing a single quote would malform the query, and the failure is invisible because the catch reports it as "no expression", which the caller reads as "abandon the conversion"), and the Statement was leaked because the try-with-resources held only the ResultSet. Both fixed, with the binding pinned by a mutation-checked test.

A third finding - that a `?` in the default expression would be parsed as a bind placeholder and fail the DDL - was checked and is not real. Driving clickhouse-jdbc 0.6.0 own `JdbcParameterizedQuery` over the emitted DDL reports zero parameters for ``DEFAULT concat(name, ?-literal)``, ``DEFAULT if(amount > 0, amount, 0)`` and the raw ternary form, against a positive control of ``INSERT ... VALUES (?, ?)`` reporting 2 and ``SELECT ... WHERE id = ?`` reporting 1 - so the probe does detect placeholders when present. ClickHouse also normalises a ternary to `if()` in `default_expression` (verified on 24.8.14 / 25.8.12 / 26.1.6), so the raw form never reaches this path. No change made.

Marked WIP: DO NOT MERGE - the merge decision is yours, not ours.

PR #1451, opened 2026-09-15
Upgrade Debezium to 3.3.0.Final to fix schema history recovery (#1450)

### LINKED ISSUE #1450: JdbcSchemaHistory cannot recover records longer than 65000 chars (Debezium 3.1.3; fixed upstream in 3.3.0 as DBZ-8979)
JdbcSchemaHistory: connector can't restart after a schema-history record exceeds 65000 chars

The embedded connector ships Debezium 3.1.3.Final, where JdbcSchemaHistory has a known bug: storeRecord splits a long record into 65000-char rows (history_data_seq = 0, 1, 2...), but recoverRecords parses each row as a standalone JSON document and never joins the parts. The first part is a truncated JSON, so restart fails with:

io.debezium.DebeziumException: com.fasterxml.jackson.core.io.JsonEOFException:
  Unexpected end-of-input in field name ... column: 65001

A record holds the full tableChanges for a table, so ~200 MySQL columns are enough to hit the limit (we hit it with a 202-column table — snapshot and streaming were fine, first restart killed the connector permanently). store.only.captured.tables.ddl=true doesn't help, one table's record is already too long.

Repro: JdbcSchemaHistory on ClickHouse (the default here), capture a ~200-column MySQL table, restart the connector.

Fixed upstream in 3.3.0.Final ([DBZ-8979](https://issues.redhat.com/browse/DBZ-8979)) — parts now share one id and get concatenated on recovery.

Proposal: bump version.debezium to >= 3.3.0.Final. Caveat: with the new write format the current history table DDL ReplacingMergeTree(record_insert_seq) ORDER BY id would collapse the parts of one record on a background merge — it needs ORDER BY (id, history_data_seq) or a different version column.

We currently run a patched recoverRecords that joins rows until the next history_data_seq = 0 before parsing. Read-side only, works with existing 3.1.3 data and the current DDL. Not upstreamable as-is, but can share if useful.

botriki:
Fixes #1450.

## Problem

`io.debezium.storage.jdbc.history.JdbcSchemaHistory` in Debezium 3.1.3 splits a schema history record into 65000-character rows on write, but `recoverRecords` parses **each row on its own** — `history_data_seq` is never consulted and the parts are never joined. Any captured table whose schema history record exceeds 65000 characters (roughly 200 columns, since the record carries the full `tableChanges` structure) therefore makes the connector unable to restart:

```
io.debezium.DebeziumException: com.fasterxml.jackson.core.io.JsonEOFException:
  Unexpected end-of-input in field name
  at [Source: REDACTED ...; line: 1, column: 65001]
```

The connector retries forever and never comes up. Debezium fixed this in 3.3.0.Final as [DBZ-8979](https://issues.redhat.com/browse/DBZ-8979). `JdbcSchemaHistory` is `final`, so this cannot be worked around inside this project — only a version bump fixes it.

## Changes

1. **`version.debezium` 3.1.3.Final → 3.3.0.Final** in the root pom and both module poms. No source change was needed; the Debezium API used here is unchanged between the two versions.

2. **Schema history table now sorts by `(id, history_data_seq)`** instead of `id`, in every config template, the Helm chart, the k8s manifests, the integration environments and the affected test. This is required by the fix rather than cosmetic: from 3.3.0 all parts of one record share the same `id` **and** the same `record_insert_seq`, so under `ReplacingMergeTree(record_insert_seq) ORDER BY id` a background merge collapses the parts of a long record into a single row and reintroduces the same failure from the write side.

3. **Explicit `log4j-1.2-api` test dependency.** Debezium 3.1.3 pulled the Log4j 1.x bridge in transitively; 3.3.0 no longer does, and 24 test classes import `org.apache.log4j`.

## Upgrade note for existing deployments

Deployments created before this change have a schema history table sorted by `id` only. After upgrading it needs to be widened, otherwise long records will be collapsed by a merge:

```sql
ALTER TABLE <db>.replicate_schema_history MODIFY ORDER BY (id, history_data_seq);
```

Rows already written by 3.1.3 give every part its own random UUID, so the 3.3 reader (which groups by `id`) cannot reassemble them. A deployment that already has an oversized record stored will need `snapshot.mode=recovery` once, after the table is fixed.

## Testing

- `mvn -DskipTests package` on JDK 17: `BUILD SUCCESS` for all three modules.
- Verified the shipped `clickhouse-debezium-embedded` jar carries the 3.3.0 `JdbcSchemaHistory` that joins rows before parsing.
- The `*IT` suites were **not** run: they need MySQL, Postgres, Oracle and ClickHouse containers, so CI is the right place for them. Opened as a draft for that reason.

Reproducer for the original bug: capture a MySQL table with ~200 columns using `schema.history.internal=io.debezium.storage.jdbc.history.JdbcSchemaHistory`, let the connector start once, then restart it.


botriki:
Can someone approve the workflow run? It stopped at `action_required`, first PR from a fork, no jobs started: https://github.com/Altinity/clickhouse-sink-connector/actions/runs/34954031079

One thing worth calling out separately, it is a breaking change for existing setups.

From 3.3.0 all parts of a split record share the same `id` and the same `record_insert_seq`. With the current default:

```
ENGINE=ReplacingMergeTree(record_insert_seq) ORDER BY id
```

a merge now collapses the parts of a long record into one row, so the bug comes back from the write side. This PR fixes the shipped templates, but existing deployments keep the old table and need:

```sql
ALTER TABLE <db>.replicate_schema_history MODIFY ORDER BY (id, history_data_seq);
```

And if someone already hit the bug: rows written by 3.1.3 have a random UUID per part, the 3.3 reader groups by `id` and cannot put them back together. One `snapshot.mode=recovery` run after fixing the table, offset is kept, no data re-snapshot.

Probably worth a release note.

Build passes on JDK 17 for all three modules, the built jar has the 3.3.0 `JdbcSchemaHistory`. I cannot run the IT suites locally.

botriki:
Pushed c4ca4d5b, one more behaviour change 3.3.0 brings. Worth knowing even if you take a different fix.

The first run failed 123 tests in the truncate suite. I compared it against your own green run on 2.11.0 (34749974281) and everything else lines up: multiple databases 28 vs 32, compound alters 8 vs 8, parallel alters 6 vs 6, sanity 2 vs 2. So truncate is the only thing my change broke. `java-tests-lightweight` is red on your last four runs too, including 2.11.0, so I left that alone.

Cause: 3.3.0 added `TRUNCATE_STATEMENT_PATTERN` to `BinlogStreamingChangeEventSource`. A `TRUNCATE TABLE` is now matched there and routed to the truncate operation instead of the DDL path. `skipped.operations` defaults to `t`, so it gets dropped. Before 3.3.0 the statement went through the DDL parser and you replicated it from there — the green log has `Applying: TRUNCATE TABLE ...` 16 times per table, mine has none, and truncate statements in the connector log went from 48 to 4.

Fix in the new commit: the connector defaults `skipped.operations` to `none` unless the user set it. Two unit tests cover the default and that an explicit user value still wins.

This hits anyone upgrading, not just this PR. If someone bumps Debezium without it, truncate replication stops with nothing in the log to explain it. Happy to change the approach if you would rather set it in the config templates, or handle the truncate operation in the sink instead of keeping it on the DDL path.

aadant:
@botriki : I do not think you can run the SQL statement on an existing table, it can only be run if you add the column at the same time. So a more complex migration path need to happen (create the new table, insert select, swap the tables), ideally automatically.

botriki:
@aadant you are right, my `ALTER` advice was wrong. `MODIFY ORDER BY` only accepts columns added by `ADD COLUMN` in the same statement, and `history_data_seq` already exists, so that statement never runs. Sorry for the noise.

On the migration path, I would not move the rows at all. Schema history is derivable from the source: `snapshot.mode=recovery` rebuilds it, keeps the stored offset and does not re-snapshot data. So recreating the table with the new sorting key and starting once in recovery is enough, and it is simpler than create / insert select / swap.

One detail that changes how urgent this is. The history read has no `FINAL`, so nothing breaks at upgrade time — the parts are separate and every row comes back. It breaks later, when a background merge collapses the rows sharing an `id` into one. A silent failure days after the upgrade is exactly what an automatic migration would be trying to prevent, but it would also move data on everyone's production for a problem most of them do not have: only deployments where a single history record exceeds 65000 chars are affected, roughly 200+ columns on one captured table, and those are already broken today.

So instead of migrating automatically I would fail loudly. On startup read the sorting key of the schema history table, and if it does not contain `history_data_seq`, stop with a message saying to recreate the table and start once with `snapshot.mode=recovery`. About twenty lines plus a unit test, no data movement, and the operator finds out at upgrade time rather than a week later.

Happy to push that as another commit here if you like the shape of it. One thing I cannot check from outside: whether the connector's own account is always allowed `SHOW CREATE TABLE` or `system.tables` — on our cluster the service account has several system tables closed. If that is not guaranteed, the check degrades to a warning in the log.

botriki:
Latest run: all four testflows jobs are green now, `java-tests-lightweight` is the only red one. Before anyone digs into it, most of it is not from this PR, and the part that is has no clean answer yet.

438 tests, 12 failures, 3 errors. I diffed it against your own run on 2.11.0 (34749974281), which fails the same job with 4 classes: `PostgresInitialDockerIT`, `PostgresInitialDockerWKeeperMapStorageIT`, `PostgresSnapshotCompletionIT`, `MySQLDemoIT`. Those four are red on both sides, so eight classes are mine.

The one that explains the shape of it:

```
TruncateTableIT.testRowsInsertedAfterTruncateSurvive:
  rows inserted after the truncate were wiped by it expected:<3> but was:<0>
```

That is my `skipped.operations=none`. With it, TRUNCATE arrives through the async data path instead of the DDL path, and loses its ordering against the rows around it — it lands after the inserts that followed it and wipes them.

Reverting does not help either, which is the actual problem:

| skipped.operations | testflows truncate suite | TruncateTableIT |
|---|---|---|
| default `t` | 123 failures | passes |
| `none` | green | fails on ordering |

So 3.3.0 breaks truncate for MySQL whichever way the flag goes; only the set of red tests changes. Before 3.3.0 the statement went through the DDL parser and was applied in order, which is why neither suite complained.

A real fix means the sink handling the truncate operation in the same order as the rows around it, rather than applying it from a different path. That is your call on the sink's semantics, not something I want to slip into a version-bump PR, so I have left the flag as it is and I am happy to go whichever way you prefer:

1. Handle op `t` in order in the sink, and keep `skipped.operations=none`.
2. Keep Debezium's default, accept that MySQL truncate replication stops at 3.3.0, and mark the testflows truncate suite accordingly.
3. Something narrower I have not thought of.

The Postgres failures in my list I have not traced individually. They may follow from the same out-of-order truncate in test fixtures, or be 3.3.0 Postgres changes — worth a second look once the truncate question is settled, since that decides whether I keep the flag at all.

PR #1459, opened 2026-09-18
test: fix the two CI failures on 2.11.0 left by #1455 and #1457

minguyen9988:
## Summary

Fixes the two remaining CI failures on the 2.11.0 branch (#1437): the java-tests-lightweight failure and the three testflows-lightweight failures (x86, hikari-pool, hikari-pool replicated). Test-only change; no connector code is touched.

### 1. AlterTableModifyColumnIT.testAlterAddPrimaryKeyAndModifyNotNull (java-tests-lightweight)

The test added in #1457 fails inside MySQL before anything reaches the connector:

    java.sql.SQLSyntaxErrorException: Multiple primary key defined
        at AlterTableModifyColumnIT.testAlterAddPrimaryKeyAndModifyNotNull(AlterTableModifyColumnIT.java:154)

Line 154 is step 3, "ALTER TABLE office ADD PRIMARY KEY (office_id), MODIFY COLUMN ..., ADD COLUMN ...". alter_ddl_modify_column.sql already creates office with "office_id INT PRIMARY KEY", so MySQL rejects the second key and the multi-clause translation the step exists to exercise (ADD PRIMARY KEY as the first clause must not leave a leading comma) never runs.

Fix: add a keyless table "branch" to the init script and run step 3 against it. The test now also asserts the result on the ClickHouse side: branch_name stays Nullable(String) (MODIFY ... NOT NULL on an existing column), branch_status is String (ADD COLUMN ... NOT NULL is honored), and the ADD PRIMARY KEY clause emitted nothing.

### 2. testflows "alter/modify column not null" and "alter/change column name to new name not null"

All 2x4 failing examples in the three testflows jobs are exactly these two features. Each example times out after 100s waiting for DESC to show "x","String" / "x2","String".

Since #1455, MODIFY/CHANGE COLUMN ... NOT NULL keeps the existing ClickHouse column Nullable (converting Nullable to non-Nullable needs a DEFAULT expression, otherwise ClickHouse rejects the ALTER with Code: 36 and the indefinitely retried DDL stalls the stream), so the column now reads Nullable(String). #1455 updated the Java unit assertions to the Nullable form but not these two testflows expectations. This aligns them and records the rationale in the feature docstrings. No other suite asserts a non-Nullable result after MODIFY/CHANGE ... NOT NULL (the plain "modify column" feature, which passed, still expects String because the translator consults the current ClickHouse nullability when NULL/NOT NULL is not given).

## Verification

- MySqlDDLParserListenerImplTest: 114 tests, 0 failures, 0 errors (JDK 17, Maven 3.9.9).
- sink-connector and sink-connector-lightweight compile including test sources.
- The integration test and the testflows suites require Docker and are left to CI on this PR.


minguyen9988:
Added a second commit (d83f5685) fixing the one remaining flaky example, so this PR now makes 2.11.0 CI deterministic end to end.

Root cause of the flake: in the original 2.11.0 run, the example "is deleted / is deleted different datatypes / check is_deleted with DECIMAL(2,1) NOT NULL" failed in only one of the four testflows-lightweight jobs (x86 replicated) and passed in the other three, failing fast (about 6.8s) at the INSERT rather than timing out. generate_sample_mysql_value() drew DECIMAL values from round(random.uniform(-(10**(p-s)), 10**(p-s)), s), but DECIMAL(p,s) admits an absolute value of at most (10**p - 1)/10**s — that is 9.9 for DECIMAL(2,1), not 10.0. Any draw at or above max minus 0.5*10**-s rounded up to exactly 10**(p-s), one unit of the last place out of range, and MySQL 8.0 strict mode rejected the INSERT with "Out of range value" (about 0.5% per run for DECIMAL(2,1), i.e. one job in four).

Fix: draw an integer number of least-significant-digit units over the exact valid range and place the decimal point by string, so every value is in range by construction; building the string from the int also avoids scientific notation (which MySQL rejects) and the default Decimal context 28-digit limit for DECIMAL(30,10). Verified over 300k draws each for DECIMAL(2,1), (30,10), (10,2), (5,0) and (1,1): zero out-of-range, zero scientific-notation. The only callers are is_deleted.py (all DECIMAL types) and multiple_databases.py (INT/VARCHAR only, unaffected).

PR #1476, opened 2026-09-22
fix(writer): engine semantics -- resolved version/sign columns, CollapsingMergeTree cancel row, relocation on stored values, TRUNCATE at its binlog position, no dropped records

minguyen9988:
Base: 2.11.0 at 5780b626 (after 1470-1474). Part of the replication-correctness audit; spec-first, failing-then-passing test and mutation check per finding; Lean BatchOrder model added for the TRUNCATE ordering.

- Sorting-key relocation is detected on stored values (getWithoutDefault), not on Connect-schema defaults: a key column moving NULL to its DEFAULT compared equal, so no tombstone was written and the old row was stranded. Spec 05.01.
- CollapsingMergeTree UPDATE stages the cancel row (sign -1) with its own addBatch; previously the before image was bound and then overwritten by the after image, so deletes and updates never collapsed. Every UPDATE is grouped once instead of twice. Specs 04.04, 05.04, 04.01.
- No record is dropped at the grouping stage; an empty batch is refused loudly. The missing-image case used to fail with an opaque NPE and retry forever.
- Kafka sink task fails when its runnable is dead; only true tombstones are skipped.
- A TRUNCATE that ClickHouse keeps refusing fails the batch instead of being reported as applied after ten retries.
- A replicated TRUNCATE change event is applied at its binlog position, against the target database name; before, TRUNCATE events grouped with DML ran in HashMap order (rows after it could be lost, rows before it resurrected) and against the source database name.
- Engine columns are bound by their resolved names: a ReplacingMergeTree whose version or is_deleted column is not literally _version or is_deleted, or a CollapsingMergeTree whose sign column is not _sign, previously never had the column bound (version 0 for every row, sign 0, deletes never collapsing). A ReplacingMergeTree that cannot be versioned or deleted from is refused.

Verification: sink-connector unit suite and lightweight unit suite green; Docker-backed ITs (TruncateTableIT, BinLogHistoryIT) rely on CI. Spec validator passes; lake build clean.


minguyen9988:
Pushed 7f35e6fa on top of a merge of the current 2.11.0 head (cb523f19, PR 1475), which resolves the conflict, and it fixes the two red CI jobs of run 35688106750. Root cause of both was the same: DbWriter.requireReplacingMergeTreeColumns (a8fca875) refused every ReplacingMergeTree target whose delete column is not a table column. That is too strict. INSERTs and UPDATEs replicate correctly to such a table, and old-style ReplacingMergeTree(ver) targets without a delete column are common: the fixtures behind DbWriterTest, DbKafkaOffsetWriterTest and DBMetadataTest.getTestGetServerTimeZone (java-tests-kafka, 3 errors, table employees.employees with delete column sign) and the pre-created public.schema_drift_* tables of PostgresSchemaDriftIT (java-tests-lightweight, 2 failures) all failed on it. Only a row that needs the delete marker is wrong on such a table. Change: the writer now logs a WARN for a missing delete column and keeps the table (a missing version column is still refused, since every row written would be wrong); PreparedStatementFieldMapper.requireDeleteColumn refuses, with IllegalStateException, a DELETE record and the sorting-key relocation tombstone for a ReplacingMergeTree target whose delete column is not in the column map (unless ignore_delete=true; replication-history mode is exempt, like requireEngineColumnPlaceholder). Also adapts PreparedStatementExecutorClearParametersTest from 2.11.0 to the ordered segment list this branch introduced. Specs 08.01 section 3.2, 05.02 section 3.1 and 04.02 section 5 updated; spec validator PASSED; 4 new mapper tests plus the rewritten writer test; mutation check: disabling requireDeleteColumn fails the DELETE and tombstone tests; full sink-connector unit suite 539 run, 0 failures, 20 Docker-environment errors identical to base.

PR #1477, opened 2026-09-22
fix(cdc): lifecycle -- unconvertible ROW records are terminal, offset FIFO survives restarts, terminal failures terminate, validated offset edits, binlog_row_image preflight

minguyen9988:
Base: 2.11.0 at 5780b626 (after 1470-1474). Part of the replication-correctness audit; spec-first, failing-then-passing test and mutation check per finding; Lean Snapshot and OffsetFifo models extended.

- An unconvertible ROW record is terminal, never acknowledged as a heartbeat. processEveryChangeRecord returned null for a ROW record whose parse failed, and handleChangeEventBatch committed that offset like a control record: the row was lost with no redelivery. Spec 01.06 and two tests had codified that behaviour; both are inverted deliberately.
- In-process engine restart (REST or the restart monitor) no longer poisons the static offset FIFO: ordered stop (engine, event thread, bounded drain, pool, FIFO reset); setup refuses while unwritten batches exist. Before, a unit handed off but never written stayed FIFO head for the life of the JVM, offsets froze while inserts continued.
- Terminal engine failures terminate the process; a slow DDL drain is not treated as terminal.
- REST /binlog and /lsn edits are validated; the stored offset is read as the newest row.
- Preflight refuses to start against a MySQL source whose binlog_row_image is not FULL (partial images make the relocation and version logic unsound).
- enable.time.adjuster=false is forced unless the user set it explicitly.
- DDL is filtered by the capture lists; disable.drop.truncate is live and scoped to table and database drops.
- Deployment defaults and dead keys cleaned up; a column-override mismatch halts; ignore_delete is announced at startup as loss by design.

Verification: lightweight unit suite green; lake build clean, 0 sorry; spec validator passes. Docker-backed ITs rely on CI.


minguyen9988:
Pushed 76439320: merged the current 2.11.0 head (40752a46, PR 1476) into the branch to resolve the conflict, and fixed the red java-tests-lightweight job. Conflict resolution (all kept-both): DbWriter's two init catch clauses, the Lean README module list, spec 10.04's verification list, and spec 07.03 (the out-of-range section from 1476 kept, this branch's enable.time.adjuster section renumbered to 3.4 with its code cross-references updated). CI root cause: this branch made disable.drop.truncate live for the first time, and DateTimeWithUserProvidedTimeZoneSchemaOnlyIT sets disable.drop.truncate=true together with enable.snapshot.ddl=true. The snapshot-phase schema bootstrap (a table drop-then-recreate Debezium emits for each captured table) was being suppressed, so the pre-created target table (Minimum_Value as DateTime, min 1970-01-01) survived instead of being replaced by the DateTime64 the snapshot recreate builds (min 1900-01-01 = DATETIME_MIN); the assertion at line 152 saw 1970-01-01 and failed. Fix: disable.drop.truncate now suppresses STREAMING drops only; snapshot-phase DDL (isSnapshotDDL) is exempt, because a snapshot recreate is schema initialisation, not a source-initiated data drop, and suppressing it would freeze a stale schema and diverge from MySQL. Spec 06.08 section 3.5 updated; new unit test DdlIgnoreRulesTest.disableDropTruncateExemptsSnapshotDdl (mutation-checked); DdlIgnoreRulesTest 7/7; spec validator PASSED; full sink-connector unit suite 541 run, 0 failures, 20 Docker-environment errors identical to base.

PR #1478, opened 2026-09-22
fix(versioning): durable _version floor across restarts, GTID path floored by commit order, tagged GTIDs, basename change, SCD2 composite keys

minguyen9988:
Base: 2.11.0 at 5780b626 (after 1470-1474). Part of the replication-correctness audit; spec-first, failing-then-passing test and mutation check per finding; Lean theorem for the floor. Everything stays inside the 2.8.0-compatible _version domain: the formula, seeds and encoding are untouched, only what feeds them changes (spec 02.06 section 6.1 states the compatibility contract).

- The version floor is seeded across restarts from a durable high-water mark (table replica_version_high_water, write-ahead horizon of 5 s of source time; fallback max(_version) over ReplacingMergeTree targets). Before, the sequence state reset to 0 on restart, so the first events after a restart could rank below rows written before it by as much as the replication lag plus about half a second, and an older row won under FINAL.
- Heartbeats and other control records no longer move the sequence state (they pinned the floor to the connector clock).
- The GTID (snowflake) path is fed the floored source timestamp, so a long transaction that commits late ranks after the shorter transaction that committed earlier.
- Tagged GTIDs are parsed from their last segment; snowflake.id=false logs a data snapshot at ERROR.
- The high-water mark is reset when the binary log basename changes.
- SCD2 history rows are closed by the whole composite primary key; the row index is bound as sequence on the Kafka Connect path.
- Specs state that record-identity de-duplication is Kafka-only and that GTID same-millisecond ties rest on the FINAL tie rule.

Verification: sink-connector unit suite 470 green (20 Docker-only errors, identical to base); lightweight unit suite green after the seam fix in the last commit; lake build clean; spec validator passes. No end-to-end upgrade test (2.10.x writes then 2.11.0 restart) could be run without Docker; it is recorded as a gap in spec 02.06.


minguyen9988:
Pushed cd0d2646: merged the current 2.11.0 head (d1663dee, which now contains #1475/#1476/#1477) into the branch to resolve the conflict. Eight conflicts. Seven were additive (kept both sides): the Lean lakefile/README, CONSTITUTION, and spec 11.03 list both BatchOrder (from #1476) and this branch's VersionFloor (the numbered proof list renumbered so BatchOrder is 13 and VersionFloor 14-16); spec 04.04 keeps #1476's collapsing/grouping bullets plus this branch's composite-PK bullet; spec 01.06 keeps #1477's terminal classifier plus this branch's heartbeat-sequence-state test. The one semantic conflict was the version-assignment hot loop of DebeziumChangeEventCapture, where both behaviours are preserved: this branch's VersionAssignment machinery (nextVersionAssignment + coverAssignedVersion, control records excluded from the sequence, rows versioned from the assignment) AND #1477's loud-failure path (rejectUnrepresentableValue before the sequence; a parser throw or a null-parsed row record becomes a terminal RecordReplicationException instead of a WARN-and-skip; the RecordReplicationException/DDLReplicationException rethrows ahead of the catch-all). NullParsedRowRecordIsTerminalTest keeps its terminal assertion, adapted to the VersionAssignment reflective signature. Verification: lightweight compile clean; the four affected unit tests 38/38; spec validator PASSED (59 specs, 519 citations); lake build completed successfully with both BatchOrder and VersionFloor and zero sorry; full sink-connector unit suite 547 run, 0 failures, 20 Docker-environment errors identical to base. CI is running on the merged head now.

PR #1479, opened 2026-09-22
fix(ddl): translation identity and types -- keyless ORDER BY, PRIMARY KEY changes loud, unsigned/SERIAL/FLOAT(M,D), spatial, CHECK, back-fill defaults, NOT NULL on MODIFY

minguyen9988:
Base: 2.11.0 at 5780b626 (after 1470-1474). Part of the replication-correctness audit; spec-first, failing-then-passing test and mutation check per finding; Lean: Clause.primaryKeyChange joins the loud set with primary_key_change_is_loud.

- A keyless CREATE TABLE is keyed by every stored column on the DDL path, never ORDER BY tuple() (matching what 1474 did on the auto-create path).
- Every unsigned integer spelling, SERIAL and FLOAT(M,D) map to types that hold the source values (they were mapped signed or to invalid output).
- The MySQL spatial family maps to String (WKB hex) on CREATE and ALTER, consistent with the writer.
- CHECK constraints (ADD CHECK, DROP CONSTRAINT) are skipped instead of echoed as invalid ClickHouse DDL.
- The ADD COLUMN back-fill matches MySQL, and literal DEFAULTs are rewritten to ClickHouse syntax.
- Sorting-key DateTime types are compared without their timezone argument, so a timezone-only restatement is not treated as a key type change.
- MODIFY/CHANGE ... NOT NULL keeps the column's existing nullability instead of widening it to Nullable.
- ADD PRIMARY KEY / DROP PRIMARY KEY that changes the row identity the replica is keyed by stops loudly with both keys and the rebuild named; a restatement of the same key set is skipped; an unknown target key keeps the previous skip behaviour. Before, both clauses were skipped unconditionally and rows the source keeps distinct collapsed under FINAL.

Verification: per-commit red/green/mutation runs; lake build clean, 0 sorry; spec validator passes (59 specs, 324 citations); lightweight unit suite run on the final head. Docker-backed ITs rely on CI.


minguyen9988:
Pushed f3da0a22: merged the current 2.11.0 head (6851ad56, now containing #1475/#1476/#1477/#1478) into the branch to resolve the conflict, and fixed the CI. Nine conflicts, all additive except spatial. MySqlDDLParserListenerImpl: both sides added independent fields, kept both. Lean lakefile/README, CONSTITUTION and spec 11.03 list BatchOrder (#1476), VersionFloor (#1478) and this branch's CreateTable. spec 07.01/07.02/06.05 keep both sides' rules and citations. spec 07.06 (spatial) is a real two-path merge: the record-schema path keeps POINT/POLYGON native (#1475), the DDL path types every spatial column as String/WKB-hex (this branch); they agree on every non-polygon type and RecordSchemaVsDdlTypeAgreementTest does not cover spatial, so the code coexists. CI fix: 2.11.0 is internally inconsistent - #1478 changed processEveryChangeRecord's last parameter to VersionAssignment but DdlIgnoreRulesTest (from #1477) still reflected long.class, throwing NoSuchMethodException in java-tests-lightweight; corrected the reflective lookup to VersionAssignment (mirroring NullParsedRowRecordIsTerminalTest). The same one-line fix should also be applied to 2.11.0 directly. Verification: spec validator PASSED (59 specs, 534 citations); lightweight compiles; DDL + terminal unit tests 175/175; full lightweight unit suite 556 run, 0 failures, 62 Testcontainers/Docker errors (no daemon here); config.yml.j2 rendered + YAML-linted.

